PR #12, opened 2018-12-16
C++-ization, phase 2

nicolasnoble:


nicolasnoble:
Only globals remaining are from PGXP and the debugger's TCP code.

PR #13, opened 2019-01-06
Gpu cleanup

nicolasnoble:


nicolasnoble:
That's enough cleanup for now.

PR #198, opened 2020-05-21
Fix building on Linux

TotalCaesar659:
It makes building on Linux go further.


nicolasnoble:
Ah, thanks :-)

There is CI on this code, but it's on a fairly narrow segment of compilers / libraries. I wouldn't mind adding more to the CI.

PR #199, opened 2020-05-21
Dev/adds mac os GitHub actions

manz:


src/core/gdb-server.h
```diff
@@ -10,7 +10,7 @@
  *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
  *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
  *   GNU General Public License for more details.                          *
- *                                                                         *
+ *                                                                         *O
```
nicolasnoble, line 13: *cough*

manz: It should be fixed, feel free to resolve the conversation. 🐣

PR #204, opened 2020-05-21
Fix building on Linux

TotalCaesar659:
It makes building on Linux go further.

### DIFF
--- a/src/gui/gui.cc
+++ b/src/gui/gui.cc
@@ -919,7 +919,7 @@ void PCSX::GUI::shellReached() {
     if (g_emulator->settings.get<PCSX::Emulator::SettingFastBoot>()) regs.pc = regs.GPR.n.ra;
 
     if (m_exeToLoad.empty()) return;
-    std::u8string filename = std::move(m_exeToLoad);
+    PCSX::u8string filename = std::move(m_exeToLoad);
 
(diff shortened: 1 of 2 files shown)


src/gui/gui.cc
```diff
@@ -919,7 +919,7 @@ void PCSX::GUI::shellReached() {
     if (g_emulator->settings.get<PCSX::Emulator::SettingFastBoot>()) regs.pc = regs.GPR.n.ra;
 
     if (m_exeToLoad.empty()) return;
-    std::string filename = std::move(m_exeToLoad);
+    std::u8string filename = std::move(m_exeToLoad);
```
nicolasnoble, line 922: This probably ought to read `PCSX::u8string`

src/gui/widgets/source.cc
```diff
@@ -48,7 +48,7 @@ void PCSX::Widgets::Source::draw(const char* title, uint32_t pc) {
         if (m_oldPath != path) {
             m_oldPath = path;
             if (!std::filesystem::exists(path)) return false;
-            std::ifstream src(path.u8string());
+            std::ifstream src(path.c_str());
```
nicolasnoble, line 51: `path.string().c_str()`

PR #219, opened 2020-05-25
Fix Linux build

TotalCaesar659:
Fix Linux build after https://github.com/grumpycoders/pcsx-redux/pull/215 and https://github.com/grumpycoders/pcsx-redux/pull/217.

### DIFF
--- a/src/gui/gui.cc
+++ b/src/gui/gui.cc
@@ -325,7 +325,7 @@ void PCSX::GUI::endFrame() {
     bool changed = false;
 
     if (m_fullscreenRender) {
-        ImTextureID texture = reinterpret_cast<ImTextureID*>(&m_offscreenTextures[m_currentTexture]);
+        ImTextureID texture = reinterpret_cast<ImTextureID*>(m_offscreenTextures[m_currentTexture]);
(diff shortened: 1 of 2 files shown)


src/gui/gui.cc
```diff
@@ -325,7 +325,7 @@ void PCSX::GUI::endFrame() {
     bool changed = false;
 
     if (m_fullscreenRender) {
-        ImTextureID texture = reinterpret_cast<ImTextureID*>(&m_offscreenTextures[m_currentTexture]);
+        ImTextureID texture = reinterpret_cast<ImTextureID*>(m_offscreenTextures[m_currentTexture]);
```
nicolasnoble, line 328: Woops ;-)

TotalCaesar659: Do you mean I need to revert this?

nicolasnoble: No, no, it was a problem from the original change you've done. That's the proper fix here.

TotalCaesar659: Yes, I haven't seen it then. Here it is some docs if you need: https://github.com/ocornut/imgui/issues/1840

PR #221, opened 2020-05-26
Openbios event refactor

JaCzekanski:
- Bash build script for macOS/Linux.
- Suppressed `warning: always_inline function might not be inlinable [-Wattributes]` warning
- Refactored events to use constants for classes, modes and flags (no event spec for now)
- **BUGFIX**: `deliverEvent` was doing an assignment instead of comparison which caused to destroy event if it wasn't fired immediately (no event was delivered since the callback is null)
```c
// was
else if ((ptr->mode = 0x1000) && ptr->handler) ptr->handler();
// changed to
else if (ptr->mode == EVENT_MODE_CALLBACK && ptr->handler)
```
After this fix I was able to boot WipeOut using openbios.

### DIFF
--- a/src/mips/.gitignore
+++ b/src/mips/.gitignore
@@ -1,0 +1,7 @@
+# Clion
+.idea/
+cmake-build-*/
+
+*.bin
+*.elf
+*.map
--- a/src/mips/.gitignore
+++ b/src/mips/.gitignore
@@ -5,3 +5,5 @@ cmake-build-*/
 *.bin
 *.elf
 *.map
+*.cpe
+*.ps-exe
--- a/src/mips/openbios/kernel/events.c
+++ b/src/mips/openbios/kernel/events.c
@@ -55,7 +55,7 @@ static int getFreeEvCBSlot(void) {
     ptr = __globals.events;
     end = (struct EventInfo*)(((char *) ptr) + __globals.eventsSize);
     while (ptr < end) {
-        if (ptr->flags == 0) return slot;
+        if (ptr->flags == EVENT_FLAG_FREE) return slot;
         ptr++;
         slot++;
     }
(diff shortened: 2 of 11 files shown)


src/mips/openbios/kernel/events.c
```diff
@@ -82,23 +82,23 @@ __attribute__((section(".ramtext"))) void deliverEvent(uint32_t class, uint32_t
     ptr = __globals.events;
     end = (struct EventInfo*)(((char *) ptr) + __globals.eventsSize);
     while (ptr < end) {
-        if ((ptr->flags == 0x2000) && (class == ptr->class) && (spec == ptr->spec)) {
-            if (ptr->mode == 0x2000) ptr->flags = 0x4000;
-            else if ((ptr->mode = 0x1000) && ptr->handler) ptr->handler();
```
nicolasnoble, line 87: Ugh. Why do I keep making these stupid mistakes. Thanks for catching it!

src/mips/openbios/kernel/events.h
```diff
@@ -26,6 +26,32 @@ SOFTWARE.
...
+    EVENT_MODE_NO_CALLBACK = 0x2000,
+};
+
+enum event_flag {
+    EVENT_FLAG_FREE       = 0x0000,
+    EVENT_FLAG_DISABLED   = 0x1000,
+    EVENT_FLAG_ENABLED    = 0x2000,
+    EVENT_FLAG_PENDING    = 0x4000,
+};
+
```
nicolasnoble, line 54: Do these belong here or rather somewhere in the "common" folder?

This might be useful for non-bios software too, right?

JaCzekanski: Right, it might be useful for the shell. 
`common/kernel/event.h` ?

src/mips/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Clion
+.idea/
+cmake-build-*/
+
+*.bin
+*.elf
+*.map
```
nicolasnoble, line 7: We may also want to add .cpe and .ps-exe here, I think.

src/mips/openbios/kernel/events.h
nicolasnoble: Sounds good to me.

PR #226, opened 2020-05-30
Rest interface work

nicolasnoble:


nicolasnoble:
That's a bit crude, but that'll do for now.

PR #234, opened 2020-06-07
Threads

JaCzekanski:
- Refactored IREG and IMASK to use IRQ_### constants
- Added `disableEvent` (THPS2 is using it)
- Changed `syscall_deliverEvent(uint32_t class, uint32_t mode)` to `syscall_deliverEvent(uint32_t class, uint32_t spec)`
- Fixed infinite loop in loadAndExec path parsing
```
// was
    while (((c = *ptr) != 0) || (c != ';')) ptr++;
// is
    while (((c = *ptr) != 0) && (c != ';')) ptr++;
```

- Changed `__globals` structure to store Process and Thread array - there is always only one process that simply points to currently active thread.
- Added `openThread`, `closeThread`, `changeThread`, `getFreTCBslot` functions
- Added `changeThreadSubFunction` syscall - **Need help there, don't know how to force GCC asm to return `v0`, hardcoded `return 1` for now**

- Rewritten findDirectoryEntryForFilename which was not able to run find .exe in subdirectories - needed for Mega Men X3, MGS. The function was looking for a subdirectory entry when already entered the subdirectory. The original function was too messy to see what was wrong there.
- Added stubs for `sysInitMemory` and few memory card functions - allows more games to boot.

Metal Gear Solid is one of the games that use PSX Threads, sadly it loops with `ChangeThread` behavior - might be memory card related.

### DIFF
--- a/src/mips/openbios/handlers/irq.c
+++ b/src/mips/openbios/handlers/irq.c
@@ -43,17 +43,17 @@ static __attribute__((section(".ramtext"))) int IRQVerifier(void) {
     // they can't be cached by the compiler, if this is what the
     // original author was thinking.
     uint32_t mask = IMASK & IREG;
-    if ((mask & 0x004) != 0) deliverEvent(EVENT_CDROM, 0x1000);
-    if ((mask & 0x200) != 0) deliverEvent(EVENT_SPU, 0x1000);
-    if ((mask & 0x002) != 0) deliverEvent(EVENT_GPU, 0x1000);
-    if ((mask & 0x400) != 0) deliverEvent(EVENT_PIO, 0x1000);
-    if ((mask & 0x100) != 0) deliverEvent(EVENT_SIO, 0x1000);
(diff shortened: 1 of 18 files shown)


src/mips/openbios/handlers/irq.c
```diff
@@ -79,49 +80,39 @@ int __attribute__((section(".ramtext"))) enqueueIrqHandler(int priority) {
...
-}
-static __attribute__((section(".ramtext"))) void T3handler(int v) {
-    if (!s_timersAutoAck[3]) return;
-    IREG = ~0x80;
+    IREG = ~IRQ_TIMER2;
     returnFromException();
 }
 
-static struct HandlerInfo s_rcntHandlers[4] = {
+static struct HandlerInfo s_rcntHandlers[3] = {
```
nicolasnoble, line 115: Looks like I really wanted to make it have 4 counters.

PR #282, opened 2020-11-02
Detect pad disconnect/reconnect then reinit

JonathanDotCel:


src/gui/gui.cc
```diff
@@ -271,6 +272,19 @@ void PCSX::GUI::startFrame() {
     glfwPollEvents();
     SDL_PumpEvents();
```
nicolasnoble, line 273: You definitely want to get rid of this `SDL_PumpEvents()` call here, otherwise events will get eaten in it, and the `SDL_PollEvent(&event)` right after will not see them.

```suggestion
```

nicolasnoble (approved):

PR #305, opened 2021-01-05
Approach for supporting Overlays

JaberwockySeamonstah:
This is my approach for how to support overlays, while still support non overlay projects.
To use an Overlay, a user would need to define in his MakeFile:

- OVERLAYSCRIPT

    This contains the name of there linker script inside there project. This linker script needs to define the overlay information and set a value for __heap_base
   

- OVERLAYSECTION

    This list contains the name of the overlay sections, that will be stripped from the executable. 
    
The make file also places the .elf and .ps-exe/.cpe file - including the overlays into a folder called "bin", which needs to exist.
Overlay files will be named "Overlay" follwed by there section name.

If OVERLAYSCRIPT is defined, the makefile defaults back to default.ld, containing the "old" definition for heapbase.

### DIFF
--- a/src/mips/common.mk
+++ b/src/mips/common.mk
@@ -43,11 +43,11 @@ LDFLAGS += $(LDFLAGS_$(BUILD))
 OBJS += $(addsuffix .o, $(basename $(SRCS)))
 
 all: dep $(BINDIR)$(TARGET).$(TYPE)
+
+$(BINDIR)$(TARGET).$(TYPE): $(BINDIR)$(TARGET).elf
 ifneq ($(strip $(BINDIR)),)
 	mkdir -p $(BINDIR)
 endif
-
-$(BINDIR)$(TARGET).$(TYPE): $(BINDIR)$(TARGET).elf
 	$(PREFIX)-objcopy $(addprefix -R , $(OVERLAYSECTION)) -O binary $< $@
 	$(foreach ovl, $(OVERLAYSECTION), $(PREFIX)-objcopy -j $(ovl) -O binary $< $(BINDIR)Overlay$(ovl);)
 
--- a/src/mips/common.mk
+++ b/src/mips/common.mk
@@ -70,7 +70,7 @@ DEPS += $(patsubst %.s,  %.dep,$(filter %.s,$(SRCS)))
 dep: $(DEPS)	
 
 makeDir:
-ifneq ($(strip $(OVERLAYSCRIPT)),)
+ifneq ($(strip $(BINDIR)),)
 	mkdir -p $(BINDIR)
 endif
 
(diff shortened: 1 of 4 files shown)


JaberwockySeamonstah:
Not sure what the error "##[error]Bash exited with code '2'." exactly means from the Bios build x.x"

nicolasnoble:
Since you've changed the paths for the generated binaries, the tests and builds that rely on a certain path start failing :-) either these need to be corrected, or an optional BINDIR variable is needed to indicate where to put the binaries.

src/mips/common.mk
```diff
@@ -3,10 +3,16 @@ BUILD ?= Release
 
 ROOTDIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
-CC = $(PREFIX)-gcc
+CC  = $(PREFIX)-gcc
+CXX = $(PREFIX)-gcc
```
nicolasnoble, line 7: I think you want g++ here.

src/mips/common.mk
```diff
@@ -36,35 +42,40 @@ LDFLAGS += $(LDFLAGS_$(BUILD))
...
+	$(CC) -g -o $(BINDIR)$(TARGET).elf $(OBJS) $(LDFLAGS)
 
 %.o: %.s
 	$(CC) $(ARCHFLAGS) -I$(ROOTDIR) -g -c -o $@ $<
 
 %.dep: %.c
 	$(CC) $(CPPFLAGS) $(CFLAGS) -M -MT $(addsuffix .o, $(basename $@)) -MF $@ $<
 
-%.dep: %.cc
+%.dep: %.cpp
```
nicolasnoble, line 59: We probably want to keep both cc and cpp, since both are valid C++ filenames.

src/mips/common.mk
```diff
@@ -36,35 +42,40 @@ LDFLAGS += $(LDFLAGS_$(BUILD))
...
 	$(CXX) $(CPPFLAGS) $(CXXFLAGS) -M -MT $(addsuffix .o, $(basename $@)) -MF $@ $<
 
 # A bit broken, but that'll do in most cases.
 %.dep: %.s
 	touch $@
 
-DEPS := $(patsubst %.cc,%.dep,$(filter %.cc,$(SRCS)))
-DEPS += $(patsubst %.c,%.dep,$(filter %.c,$(SRCS)))
-DEPS += $(patsubst %.s,%.dep,$(filter %.s,$(SRCS)))
+DEPS := $(patsubst %.cpp, %.dep,$(filter %.cpp,$(SRCS)))
```
nicolasnoble, line 66: Same here, we'd want to keep both cc and cpp.

nicolasnoble (approved):

PR #318, opened 2021-01-11
[GUI] Added control rebinding

wheremyfoodat:
Details:
- Minor code cleanup
- Control rebinding
- GUI menu for configuring keyboard/joypad controls
- Serialization of current GUI theme
- Serialization of current control bindings
- Fixed some erroneous JIT comments


src/gui/gui.cc
```diff
@@ -263,6 +266,7 @@ void PCSX::GUI::saveCfg() {
     j["SPU"] = PCSX::g_emulator->m_spu->getCfg();
     j["emulator"] = PCSX::g_emulator->settings.serialize();
     j["gui"] = settings.serialize();
+    j["pad"] = PCSX::g_emulator->m_pad1->getCfg();
```
nicolasnoble, line 269: let's maybe plan for the two pads by shoving this into a json array.

src/core/pad.h
```diff
@@ -54,7 +71,29 @@ class PAD {
...
+    typedef Setting<int, TYPESTRING("Pad1Square"), GLFW_KEY_Z> Pad1Square;
+    typedef Setting<int, TYPESTRING("Pad1Circle"), GLFW_KEY_D> Pad1Circle;
+    typedef Setting<int, TYPESTRING("Pad1Select"), GLFW_KEY_BACKSPACE> Pad1Select;
+    typedef Setting<int, TYPESTRING("Pad1Sstart"), GLFW_KEY_ENTER> Pad1Start;
+    typedef Setting<int, TYPESTRING("Pad1L1"), GLFW_KEY_Q> Pad1L1;
+    typedef Setting<int, TYPESTRING("Pad1L2"), GLFW_KEY_A> Pad1L2;
+    typedef Setting<int, TYPESTRING("Pad1R1"), GLFW_KEY_R> Pad1R1;
+    typedef Setting<int, TYPESTRING("Pad1R2"), GLFW_KEY_F> Pad1R2;
+
+    static Settings<Pad1Up, Pad1Right, Pad1Down, Pad1Left, Pad1Cross, Pad1Triangle, Pad1Square, Pad1Circle, Pad1Select, Pad1Start, Pad1L1, Pad1L2, Pad1R1, Pad1R2> settings;
```
nicolasnoble, line 96: These probably shouldn't be statics, otherwise they'll leak into the other pads's configuration, unless we want to expand the whole thing into a single configuration for both pads?

src/core/pad.h
```diff
@@ -20,26 +20,43 @@
...
     PAD(pad_t pad);
     ~PAD();
     void init();
     void shutdown();
+    static void updateBinding(GLFWwindow* window, int key, int scancode, int action, int mods); // Actually update the binding for the button set to be configured
     unsigned char startPoll();
     unsigned char poll(unsigned char);
+    
+    json getCfg();
+    void setCfg(const json &j);
```
nicolasnoble, line 48: There's probably a chicken-and-egg issue here when the configuration hasn't been created yet. We may want to have a "reset" method to ensure there's default values set in.

wheremyfoodat:
Oh god L2/R2 seem to be axes instead of buttons.

PR #402, opened 2021-03-01
Add per-game savestate slots alongside the "global" savestate slot.

wheremyfoodat:
This way you can use the per-game savestates for regular user stuff and the global state if you want to load arbitrary save data.
BIOS/Shell savestates generate a file called BIOS.sstate
Games use (gamecode).sstate or (filename).sstate if there's no gamecode available
Global states use Global.sstate.

Ideally multiple slots should be easy to add too later, with some fmt::format magic


nicolasnoble:
Several general comments, which is probably the right time to clean things up:

- there is already a global iso filename, and we don't want to duplicate that information; but it may be a good idea to upgrade the current variable:
https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/plugins.cc#L33
(in fact, this whole file needs to go)

- the general accessor needs to be using https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/plugins.cc#L678-L684 (but it's okay to move it to, say, cdiso.cc, which may be where it belongs)

- the reason for this is that one can specify the iso from the command line, so there's more than one way to set the iso filename (and some duplicated and redundant, as I know realize, woops):
https://github.com/grumpycoders/pcsx-redux/blob/main/src/main/main.cc#L243
https://github.com/grumpycoders/pcsx-redux/blob/main/src/gui/gui.cc#L342
https://github.com/grumpycoders/pcsx-redux/blob/main/src/gui/gui.cc#L755
https://github.com/grumpycoders/pcsx-redux/blob/main/src/gui/gui.cc#L1330

The same goes with the "exe filename".

src/gui/gui.h
```diff
@@ -162,6 +162,7 @@ class GUI final {
 
     float m_renderRatio = 3.0f / 4.0f;
     bool &m_fullscreen = {settings.get<Fullscreen>().value};
+    std::string isoFilename;
```
nicolasnoble, line 165: If you're adding new member variables, please use the m_ prefix :-)

Also, as described in the general comment, this may not be the right location for this.

And maybe we just want to use an std::filename::path for it?

src/gui/gui.cc
```diff
@@ -560,12 +560,71 @@
                     std::ofstream schema("sstate.proto");
                     SaveStates::ProtoFile::dumpSchema(schema);
                 }
-                if (ImGui::MenuItem(_("Save state"))) {
-                    zstr::ofstream save("sstate", std::ios::binary);
+
+                if (ImGui::BeginMenu(_("Save state slots"))) {
+                    for (auto i = 1; i < 10; i++) {
+                        const auto str = _("Slot ") + std::to_string (i);
```
nicolasnoble, line 566: `const auto str = fmt::format(_("Slot {}"), i);`

PR #414, opened 2021-03-03
CDROM command logging

wheremyfoodat:


src/core/cdrom.cc
```diff
@@ -1567,6 +1570,17 @@ class CDRomImpl : public PCSX::CDRom {
         StopCdda();
         lidSeekInterrupt();
     }
+
+    void logCDROM (int command) {
+        if (command >= 0x100) return; // a hack to fight the hacks in this file... yikes
```
nicolasnoble, line 1575: We may want to instead have a `bool delayed = command & 0x100` and add a "[DELAYED]" keyword in the output. Also `switch (command & 0xff)`

src/gui/widgets/log.cc
```diff
@@ -41,6 +42,8 @@ void PCSX::Widgets::Log::draw(const char* title) {
     }
     ImGui::Checkbox(_("Follow"), &m_follow);
     ImGui::SameLine();
+    ImGui::Checkbox(_("Log CD-ROM"), &PCSX::g_emulator->settings.get<PCSX::Emulator::SettingLoggingCDROM>().value);
```
nicolasnoble, line 45: Hmm.... I'm wondering if this is the appropriate place for this checkbox...

Maybe a menu?

src/core/cdrom.cc
wheremyfoodat: Ohhhhhhhhh I really like this idea actually

nicolasnoble (approved):

PR #422, opened 2021-03-06
Revert "Removing std::tuple and use simplified implementation."

nicolasnoble:
This reverts commit 51353c28dda208929ce17ea31b225ee63966b8b2, #406.


nicolasnoble:
The change causes a significant performance decrease as std::get is much straightforward than what I came up with.

PR #440, opened 2021-03-18
[OpenBIOS] Patch to fix R-Types (relies on null pointer dereferences)

wheremyfoodat:


nicolasnoble:
So... you really don't want to try and find out why the current code that does the same exact thing isn't working ? :P

src/mips/openbios/main/main.c
```diff
@@ -250,6 +250,17 @@ static void zeroUserMemoryUntilStack() {
     fastMemset((void *)0xa0010000, 0, end - 0x10000);
 }
 
+// Horror story #2314.21
+// The original BIOS writes garbage to the first few bytes of kernel RAM, 
+// However, r-types accidentally dereferences a NULL pointer, and depeending
```
VelocityRa, line 255:
```suggestion
// However, R-Types accidentally dereferences a NULL pointer, and depending
```
:3

PR #443, opened 2021-03-19
Fix CPU bugs

wheremyfoodat:


wheremyfoodat:
Amidog in the interpreter has a regression when the link in BGEZAL/BLTZAL is not delayed, but in the JIT it breaks if they are delayed in the JALR test...

what

wheremyfoodat:
Closing because it's outdated, only DIV fixes should be re-implemented in the future

PR #520, opened 2021-06-13
Reset window size on invalid config values

Kelpsy:
This fixes a problem that occurred when the config values for the window size were corrupted on exit, for example by not getting the correct values back from `glfwGetWindowSize` while minimized.

Before (window size previously set to 0):
bug

After the fix:
fixed


wheremyfoodat (approved):
Seems pretty good. I've been bitten by this bug before on Windows and it... wasn't very fun to find out.

nicolasnoble:
I've never been bitten by it, but I can believe it being an issue.

PR #548, opened 2021-07-31
Delete unused VS filter.

wheremyfoodat:


nicolasnoble (approved):

nicolasnoble (commented):
But, the source one just below...? :-P

wheremyfoodat:
> But, the source one just below...? :-P

It is used for ix86.cc

PR #591, opened 2021-08-31
Trying to add code climate.

nicolasnoble:


nicolasnoble:
Won't work. Code climate doesn't understand llvm.

Issue #502, opened 2021-06-07
File Browser - Date


danhans42:
Not sure if this is a bug as such..

The file browser seems to show the date/time a file was created under Date/Time. 
Cheeky ask - would it be possible to change this to the modified time?

Build info : https://install.appcenter.ms/orgs/grumpycoders/apps/pcsx-redux-linux64/distribution_groups/public
OS:  Linux Mint 20 Cinnamon (AMD64)

danhans42:
![sshot](https://user-images.githubusercontent.com/15861619/121049921-1119b100-c7b0-11eb-9f7f-ab79fc773c06.png)

nicolasnoble:
So unfortunately, the timestamp I'm getting is supposed to be [last write time](https://en.cppreference.com/w/cpp/filesystem/last_write_time). I think it means there's a bug somewhere in the standard library?

nicolasnoble:
Should've been fixed :)

PR #616, opened 2021-09-25
Fix timestamps in GUI file dialogs

torresam:
Addresses #502 

Verified on my Windows 10 machine that:
1) The timestamps were incorrect prior to the change reproducing #502 
2) And that they were correct after this change
image

### DIFF
--- a/src/core/system.h
+++ b/src/core/system.h
@@ -42,14 +42,14 @@ enum class LogClass : unsigned;
 typedef decltype(std::filesystem::path().u8string()) u8string;
 #define MAKEU8(x) reinterpret_cast<const decltype(PCSX::u8string::value_type()) *>(x)
 
+
 // another hack, until C++-20 properly gets std::chrono::clock_cast
-template<typename DstTP, typename SrcTP, typename DstClk = typename DstTP::clock, typename SrcClk = typename SrcTP::clock>
+template <typename DstTP, typename SrcTP, typename DstClk = typename DstTP::clock,
+          typename SrcClk = typename SrcTP::clock>
 DstTP ClockCast(const SrcTP tp) {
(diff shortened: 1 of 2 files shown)


nicolasnoble (dismissed):
Thanks!

nicolasnoble (dismissed):

nicolasnoble:
Hmmm, it sounds like Linux / MacOS builds aren't happy with it... `std::chrono::clock_cast` is supposed to be C++20, but maybe the libc++ there didn't get this feature yet? I'll try and investigate.

torresam:
> Hmmm, it sounds like Linux / MacOS builds aren't happy with it... `std::chrono::clock_cast` is supposed to be C++20, but maybe the libc++ there didn't get this feature yet? I'll try and investigate.

Hmm it looks like it's not quite supported in libc++ yet 
https://libcxx.llvm.org/Status/Cxx20.html#cxx20-status

Should be P0355R7 I believe

nicolasnoble:
Goddammit, I tried to be clever and shoe-horn in an implementation for clock_cast, but as much as Linux tolerates my implementation, MacOS says nay.

nicolasnoble:
Yeah at this point, I've exhausted what I can think of to make MacOS' compiler happy.

If someone has a Mac machine and is willing to have at it, that'd be great.

torresam:
I tested commit I just pushed on a Mac and it compiles, should work on all platforms now.
```
Software:
    System Software Overview:
      System Version: macOS 11.3.1 (20E241)
      Kernel Version: Darwin 20.4.0
```

nicolasnoble:
How... did it even work elsewhere...?

nicolasnoble (approved):

nicolasnoble:
Thanks again :)

PR #630, opened 2021-10-04
Increase bios_size to 1MB

ABelliqueux:
:p


wheremyfoodat:
@ABelliqueux
Uhhh you'd need to adjust the memory handlers and recompiler to work with 1MB BIOSes too. And you'd also want to detect a 1MB BIOS before treating a BIOS as such.

nicolasnoble:
There are lots of other places this would need to be updated unfortunately :)

(this doesn't pass basic tests because of a buffer overflow).

PR #663, opened 2021-11-07
mips: Separate overlays into separate targets

cuckydev:
This modifies mips' common.mk to split overlays into separate targets.
This solved an issue in my own project where the overlays were used as a dependency later in the makefile, which would be dropped if `-jX` were used.


nicolasnoble (approved):

nicolasnoble:
Good catch.

PR #669, opened 2021-11-21
Implements capture/decode buffers for CD audio left/right.

MarianoDiMartino:
- Updates a temporary buffer in xa.cc for CD audio left/right, which will be written to the capture buffer in spu.cc.
- Setting can be updated by the 'Capture/decode buffer IRQ' option in the SPU config. Setting will also be updated on hard reset.
- Contains several locks to assure thread-safe handling of SPU RAM.

### DIFF
--- a/src/core/psxemulator.cc
+++ b/src/core/psxemulator.cc
@@ -94,7 +94,7 @@ void PCSX::Emulator::EmuReset() {
     m_cheats->FreeCheatSearchResults();
     m_cheats->FreeCheatSearchMem();
     m_psxMem->psxMemReset();
-
+    m_spu->resetCaptureBuffer();
     m_psxCpu->psxReset();
     m_gpu->clearVRAM();
     m_pads->shutdown();
--- a/src/spu/interface.h
+++ b/src/spu/interface.h
@@ -92,6 +92,7 @@ class impl final : public SPUInterface {
 
     // spu
     void MainThread();
+    void writeCaptureBufferCD(int numbSamples);
     void SetupStreams();
     void RemoveStreams();
     void SetupThread();
@@ -131,6 +132,7 @@ class impl final : public SPUInterface {
 
     // psx buffer / addresses
     uint16_t regArea[10000];
+    // Note that SPU ram is a uint16_t, so total size is 512KB.
     uint16_t spuMem[256 * 1024];
     uint8_t *spuMemC;
     uint8_t *pSpuIrq = 0;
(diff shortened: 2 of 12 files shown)


src/core/psxemulator.cc
```diff
@@ -94,7 +94,7 @@ void PCSX::Emulator::EmuReset() {
     m_cheats->FreeCheatSearchResults();
     m_cheats->FreeCheatSearchMem();
     m_psxMem->psxMemReset();
-
+    m_spu->resetCaptureBuffer();
```
nicolasnoble, line 97: This may only be valid for a hard reset, but I don't think it matters that much.

src/spu/interface.h
```diff
@@ -134,6 +137,22 @@ class impl final : public SPUInterface {
     uint8_t *pSpuBuffer;
     uint8_t *pMixIrq = 0;
 
+
+    struct CaptureBuffer {
+        static const int CB_SIZE = 1024 * 8;
+        // These buffers have to be large enough to allow the CD-XA to stream in enough data.
+        uint16_t CDCapLeft[CB_SIZE] = {0};
```
nicolasnoble, line 144: So I think those would require to be part of the savestate to be properly migrated over during, well, a savestate. Check out spu/freeze.cc and core/sstate.h around line 116 for more on this. It might be a little bit intimidating to modify, so let me know if you need help for this.

src/spu/spu.cc
```diff
@@ -805,6 +827,8 @@ void PCSX::SPU::impl::playADPCMchannel(xa_decode_t *xap) {
 
 long PCSX::SPU::impl::init(void) {
     spuMemC = (uint8_t *)spuMem;  // just small setup
+    captureBuffer = new CaptureBuffer();
```
nicolasnoble, line 830: We probably want to delete this one around shutdown :)

That being said, why not simply making it part of the class itself instead of allocating it?

src/core/psxemulator.cc
MarianoDiMartino: Can also put this for soft reset too.

nicolasnoble: Either way is fine, really. This buffer ought to be reset pretty quickly, and I don't think we're properly resetting the SPU ram on reset at the moment anyway. Still a lot of cleanup to do :D

src/spu/spu.cc
MarianoDiMartino: Not a fan of putting all these arrays on the stack. But sure :)

nicolasnoble: Oh but that's not the stack, right? The SPU object is created as a unique_ptr from the main emulator code. It's going to be heap allocated.

MarianoDiMartino: Ahh, didn't notice that!

MarianoDiMartino:
Done with the capture buffer. Also implemented voice 1 and voice 3 writes and fixed some minor issues.
Feel free to review and merge.

nicolasnoble (approved):

PR #702, opened 2021-12-31
Dynarec Disassembler GUI Widget using Capstone

caldog20:
Capstone disassembler used to disassemble the code buffer in dynarec CPUs. Option is only available in Menu when running on a Dynarec CPU. Supports live disassembly as well as writing to output file. Current code will support x86_32/64 and aarch64.


wheremyfoodat (dismissed):

wheremyfoodat (approved):

wheremyfoodat (dismissed):

src/gui/gui.cc
```diff
@@ -863,6 +863,8 @@ void PCSX::GUI::endFrame() {
                 }
                 ImGui::MenuItem(_("Show Registers"), nullptr, &m_registers.m_show);
                 ImGui::MenuItem(_("Show Assembly"), nullptr, &m_assembly.m_show);
+                if (PCSX::g_emulator->m_psxCpu->isDynarec())
+                    ImGui::MenuItem(_("Show DynaRec Disassembly"), nullptr, &m_disassembly.m_show);
```
nicolasnoble, line 867: We probably want an else statement here to show the same string in a disabled text fashion.

src/gui/gui.cc
```diff
@@ -1060,6 +1062,10 @@ void PCSX::GUI::endFrame() {
                         _("Assembly"));
     }
 
+    if (m_disassembly.m_show) {
```
nicolasnoble, line 1065: Re-add a check here for the CPU type, in case we eventually add the boolean to the configuration file.

src/gui/widgets/dynarec_disassembly.cc
```diff
@@ -0,0 +1,206 @@
...
+
+#include <capstone/capstone.h>
+
+#include <cinttypes>
+#include <fstream>
+
+#include "gui/gui.h"
+#include "imgui.h"
+
+int PCSX::Widgets::Disassembly::writeFile() {
```
nicolasnoble, line 30: Two remarks here:

(1) This function returns either 0 or -1, which is essentially a boolean
(2) The return value is never checked against

(1) means we probably want to change to either a boolean or an enum (success / failure), and (2) potentially means we actually don't really care about returning anything,so, void.

src/gui/gui.cc
caldog20: The goal here was to prevent it from appearing in the debug menu if not running on a dynarec CPU. Are you wanting this to still show the menu item but as disabled and greyed out?

src/gui/widgets/dynarec_disassembly.cc
caldog20: Good point. I initially had the Disassemble function return an enum and would show an error popup depending on what enum result returned. I took this out because I thought it may have been a bit over-engineered. I slimmed this down to messages in the Logs and only a popup if it was a critical error. 

I will either fix this to return a bool and check on it, or make it void and have it spit a message to logging.

src/gui/gui.cc
caldog20: Let me know if you don't like the && , and I will change it to a nested if-statement

wheremyfoodat:
dynarec_disassembly.cc line 50: Unneeded `return`

src/gui/gui.cc
nicolasnoble: That's correct. Having a consistent UI is important. Not displaying a menu item under certain circumstances is bad UX because it confuses users when it happens. It's better to display the same entry, disabled, to convey to users they'd be able to use this under different settings. Maybe even add a hover text to explain why it's disabled and what to do to enable it.

src/gui/gui.cc
nicolasnoble: && is perfectly fine.

nicolasnoble (dismissed):

src/gui/widgets/dynarec_disassembly.cc
```diff
@@ -0,0 +1,204 @@
...
+
+    // Get pointer to code buffer along with size of buffer
+    const uint8_t* buffer = PCSX::g_emulator->m_psxCpu->getBufferPtr();
+    const size_t bufferSize = PCSX::g_emulator->m_psxCpu->getBufferSize();
+    // Check to ensure code buffer pointer is not null and size is not 0
+    if (buffer == nullptr) {
+        PCSX::g_system->printf("Disassembler Error: nullpointer to code buffer.\n");
+        m_showError = true;
+        return -1;
+    } else if (bufferSize <= 0) {
```
nicolasnoble, line 167: This can never happen: bufferSize is not a signed value.

nicolasnoble (approved):

PR #705, opened 2022-01-03
Add Memory Card Manager

wheremyfoodat:


src/core/sio.cc
```diff
@@ -809,31 +787,83 @@ void PCSX::SIO::GetMcdBlockInfo(int mcd, int block, McdBlock *Info) {
...
+        frame[i] = 0;
+    }
+    frame[0x7f] = 0xa0; // xor checksum of frame
+}
+// Back up the entire memory card to a file
+// mcd: The memory card to back up (1 or 2)
+void PCSX::SIO::SaveMcd(int mcd) {
+    const auto data = GetMcdData(mcd);
+    
+    if (mcd == 1) {
```
nicolasnoble, line 862: switch?

src/core/sio.cc
```diff
@@ -809,31 +787,83 @@ void PCSX::SIO::GetMcdBlockInfo(int mcd, int block, McdBlock *Info) {
...
+    // Check if the block is marked as free in the directory frame and adjust the name/filename if so
+    if (Info->Flags == 0xa0) {
+        std::strcpy(Info->Title, "Free Block");
+        std::strcpy(Info->Name, "Empty File");
+        Info->Filesize = 0;
+    }
+}
 
-    Info->Flags = *ptr;
+char *PCSX::SIO::GetMcdData(int mcd) {
```
nicolasnoble, line 827: Maybe `mcd` should be of an enum type.

wheremyfoodat: Does it really make sense to use something like
```cpp
enum class MemoryCards {
    Card1, Card2, ...
};
```
When that would more or less be the exact same thing as using an integer but with more steps? An argument could also be made about `int` allowing us to support an arbitrary amount of memory cards eventually, but I don't think that's really useful.

nicolasnoble: The "more steps" is sort of precisely the point. Well, there's two reasons. One is to improve readability of calls into the API, without having to look at the prototype or the comments, in order to avoid calls in an obtuse way like

```c
doSomethingWithMcd(0, 12, 52, 28);
```

and have something a bit better like

```c
doSomethingWithMcd(MCD::Slot::S1, MCD::Command::XYZ, 52, 28);
```

It's okay to try and reduce the number of letters to type tho, I agree on this.

The second reason is to avoid basic errors like putting the arguments in the wrong order, by forcing a stronger C++ type for each argument as much as possible.

nicolasnoble (approved):

PR #710, opened 2022-01-05
[Interpreter] Overflow exceptions should fire even if dest == $zero

wheremyfoodat:


nicolasnoble (approved):
Maybe we should add a test for this one...

PR #712, opened 2022-01-11
SPU Code Refactoring - breaking up fat main loop - removing waiting c…

MrAshFar:
- Breaking up the monolithic `MainThread `method into manageable methods (more scalable)   
- Code Refactoring 
- Decreasing the probability of losing voice data by removing waiting cycles (`while `loop at `feedStreamData`)
- Adding a temporary queue in the `circular.h `to decrease voice data loss and **remove exception throwing**

### DIFF
--- a/src/spu/interface.h
+++ b/src/spu/interface.h
@@ -91,8 +91,37 @@ class impl final : public SPUInterface {
     static const size_t NSSIZE = 45;
 
     // spu
+    class main_t_variables {
+      public:
+        int s_1, s_2, fa, ns;
+        uint8_t *start;
+        unsigned int nSample;
+        int ch, predict_nr, shift_factor, flags, d, s;
+        int bIRQReturn = 0;
+        int voldiv = 0;
+        int32_t tmpCapVoice1Index = 0;
+        int32_t tmpCapVoice3Index = 0;
+        PCSX::SPU::SPUCHAN *pChannel;
+        void reset() {
+            s_1 = s_2 = fa = ns = 0;
+            start = nullptr;
+            nSample = ch = predict_nr = shift_factor = flags = d = s = bIRQReturn = voldiv = 0;
+            tmpCapVoice1Index = 0;
+            tmpCapVoice3Index = 0;
+            pChannel = nullptr;
+        }
+    } _vars;
     void MainThread();
     void writeCaptureBufferCD(int numbSamples);
+    void mixAllChannels();
+    void main_t_pMixIrq();
+    void main_t_FeedStreamData();
+    void main_t_IrqCheck();
+    void main_t_FlagHandler();
+    void main_t_StopSign();
+    void main_t_SBPos_28();
+    void main_t_get_noise_or_sample_mixedSample();
+    void main_t_fmod_freq_channel();
     void SetupStreams();
     void RemoveStreams();
     void SetupThread();
--- a/src/spu/interface.h
+++ b/src/spu/interface.h
@@ -91,7 +91,7 @@ class impl final : public SPUInterface {
     static const size_t NSSIZE = 45;
 
     // spu
-    class main_t_variables {
+    struct MainThreadVariables {
       public:
         int s_1, s_2, fa, ns;
         uint8_t *start;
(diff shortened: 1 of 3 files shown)


src/spu/interface.h
```diff
@@ -91,8 +91,37 @@ class impl final : public SPUInterface {
...
+        PCSX::SPU::SPUCHAN *pChannel;
+        void reset() {
+            s_1 = s_2 = fa = ns = 0;
+            start = nullptr;
+            nSample = ch = predict_nr = shift_factor = flags = d = s = bIRQReturn = voldiv = 0;
+            tmpCapVoice1Index = 0;
+            tmpCapVoice3Index = 0;
+            pChannel = nullptr;
+        }
+    } _vars;
```
nicolasnoble, line 113: "Each identifier that begins with an underscore is reserved to the implementation for use as a name in the global namespace." (ISO standard)

src/spu/interface.h
```diff
@@ -91,8 +91,37 @@ class impl final : public SPUInterface {
     static const size_t NSSIZE = 45;
 
     // spu
+    class main_t_variables {
```
nicolasnoble, line 94: What's the reasoning behind shoveling all of those stack variables into a member that'll be allocated on the heap? The compiler will no longer be able to hoist those into local registers, and this makes the code less readable overall.

Also the reset function is called exactly once during the whole execution, making it fairly pointless, while mixing self initialization of some variables in, but not others.

src/spu/spu.cc
```diff
@@ -896,6 +870,7 @@ void PCSX::SPU::impl::wipeChannels() {
 ////////////////////////////////////////////////////////////////////////
 // SETUPTIMER: init of certain buffers and threads/timers
 ////////////////////////////////////////////////////////////////////////
+std::future<int> main_t_rst;
```
nicolasnoble, line 873: Please don't use globals.

src/spu/spu.cc
```diff
@@ -907,8 +882,12 @@ void PCSX::SPU::impl::SetupThread() {
     bEndThread = 0;  // init thread vars
     bThreadEnded = 0;
     bSpuInit = 1;  // flag: we are inited
-
-    hMainThread = std::thread([this]() { MainThread(); });
+    // Threads are expensive and causes Thread init and switch overheads
+    // hMainThread = std::thread([this]() { MainThread(); });
+    main_t_rst = std::async([this]() {
```
nicolasnoble, line 887: If anything, std::async will also set up a thread to function properly. This only obfuscates the fact.

src/support/circular.h
```diff
@@ -41,10 +42,28 @@ class Circular {
...
+
+    std::queue<std::tuple<const T*, size_t>> tempQueue;
     bool enqueue(const T* data, size_t N) {
-        if (N > BUFFER_SIZE) {
-            throw std::runtime_error("Trying to enqueue too much data");
+        if (tempQueue.size()) {
+            // throw std::runtime_error("Trying to enqueue too much data");
+            tempQueue.push(std::make_tuple(data, N));
+            if (std::get<1>(tempQueue.front()) < BUFFER_SIZE) {
+                do {
```
nicolasnoble, line 52: The whole point here is to *not stall* for too long (see the call to `wait_for` a bit later), and take advantage of a tight streamline of data between the components to reduce audio lag. This breaks those assumptions for no effective result.

src/spu/interface.h
```diff
@@ -91,8 +91,37 @@ class impl final : public SPUInterface {
...
+            start = nullptr;
+            nSample = ch = predict_nr = shift_factor = flags = d = s = bIRQReturn = voldiv = 0;
+            tmpCapVoice1Index = 0;
+            tmpCapVoice3Index = 0;
+            pChannel = nullptr;
+        }
+    } _vars;
     void MainThread();
     void writeCaptureBufferCD(int numbSamples);
+    void mixAllChannels();
```
nicolasnoble, line 116: The naming here is a bit all over the place.

- C++ allows for namespacing, there's no need to add more prefixes.
- Some of those names aren't self explanatory, like SBPos_28.
- The general naming convention of the project is dromedaryCases for method names; leftovers code from WayBefore still needs to be converted / cleaned up.

src/spu/spu.cc
nicolasnoble, line 887: I just realized that this `std::future` is never being `get` anywhere, which probably means the code will never run...

src/spu/interface.h
MrAshFar: **_What's the reasoning behind shoveling all of those stack variables into a member that'll be allocated on the heap?_**
The main reason was that, for breaking the main thread into manageable methods, there are 2 options:

1. Passing local variables to calling functions:
 + it's faster
 - it's less readable
 - it's less scalable 

2. Hosting variables in a shared memory address:
  + common playground (shared memory address)
  + more scaleable and readable
  + more OOP (taking benefit of C++ features)
  - slower than the stack memory
so its a trade-off between scalability and speed

**_Also the reset function is called exactly once during the whole execution, making it fairly pointless_**

in some cases where we need to restart the main thread such as encountering an exception or malfunction

src/spu/interface.h
MrAshFar: ok, I'll fix it

src/spu/spu.cc
MrAshFar: I made `main_t_rst` to make it run concurrently. otherwise, it will run sequentially which Emu will hang. because the future is missed when it's not global. 

It's better to roll it back to previous code, you are fine

src/spu/interface.h
MrAshFar: Ok, I will fix it.

src/support/circular.h
MrAshFar: I will check it more. However, throwing an exception for the shortage of audio buffer is not a good idea.

src/spu/spu.cc
MrAshFar: I made the` future` a global variable to preserve its value otherwise if the value miss, the Emu hangs. because the program waits till the end of the endless while loop.

src/spu/spu.cc
MrAshFar: ok, I will fix it

src/spu/interface.h
MrAshFar: I found a better way that satisfies both 2 options. that is, using a `struct` instead of a `class`, instantiating struct as a local variable, and passing the `struct`'s ref into the methods.

lorn10:
Just out of curiosity, was this closed because of #724? 

Most likely it would make really sense to include these amazing "peops spu 1.10" improvements into a possible rework of the SPU code in PCSX-Redux. :wink:

PR #718, opened 2022-02-02
Add SIO1 Server + cdrom adjustments to support Unirom/NotPSXSerial

johnbaumann:
Created an SIO1 server based around gdb-server and web-server implementations, 

Implemented some very basic, rough, and unirom specific SIO1 callbacks and emulation.

Added some error checking around logging cdrom commands to prevent exceptions when booting Unirom.

Added some logic for clearing the Index0 DATA FIFO flag


src/gui/gui.cc
```diff
@@ -1373,6 +1374,20 @@ query using a REST api. See the wiki for details.
...
+            if (debugSettings.get<Emulator::DebugSettings::SIO1Server>()) {
+                g_emulator->m_sio1Server->startServer(&g_emulator->m_loop,
+                                                     debugSettings.get<Emulator::DebugSettings::SIO1ServerPort>());
+            } else {
+                g_emulator->m_sio1Server->stopServer();
+            }
+        }
+        ShowHelpMarker(_(R"(This will activate a tcp server, that will
+relay information between tcp and sio1.
+See the wiki for details.)"));
```
nicolasnoble, line 1388: image

johnbaumann: That last tcp should say redux, derp.

johnbaumann: Also, note to self, add wiki entry.

nicolasnoble (approved):

PR #720, opened 2022-02-02
Simplify some JIT code, prevent potential crashes.

wheremyfoodat:


nicolasnoble:
You've noticed this doesn't pass the unit tests by the way, @wheremyfoodat ?

PR #731, opened 2022-02-10
Fix R_MIPS_GPREL16 by setting the correction section attributes.

paulsapps:
And also fix bss and .sbss symbol offsets.


paulsapps:
This should be more correct than it was but is likely wrong (over) alignment for structures and arrays might also be wrong. Investigating how to handle those correctly...

paulsapps:
just to confirm this should now be correct in all known cases

PR #733, opened 2022-02-12
Added Memory Observer.

twevs:
Accessed via Debug > Show Memory Observer, provides some delta-over-time search functionality.

### DIFF
--- a/src/gui/widgets/memory_observer.cc
+++ b/src/gui/widgets/memory_observer.cc
@@ -109,8 +109,13 @@ void PCSX::Widgets::MemoryObserver::draw(const char* title) {
 
         std::erase_if(m_AddressValuePairs, doesntMatchCriterion);
 
-        for (auto& addressValuePair : m_AddressValuePairs) {
-            addressValuePair.scannedValue = getMemValue(addressValuePair.address, memData, memSize, memBase, stride);
+        if (m_AddressValuePairs.empty()) {
+            m_scanType = ScanType::ExactValue;
+        } else {
+            for (auto& addressValuePair : m_AddressValuePairs) {
+                addressValuePair.scannedValue =
+                    getMemValue(addressValuePair.address, memData, memSize, memBase, stride);
+            }
         }
     }
 
--- a/src/gui/widgets/memory_observer.cc
+++ b/src/gui/widgets/memory_observer.cc
@@ -1,5 +1,5 @@
 /***************************************************************************
- *   Copyright (C) 2019 PCSX-Redux authors                                 *
+ *   Copyright (C) 2022 PCSX-Redux authors                                 *
  *                                                                         *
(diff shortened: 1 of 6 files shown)


src/gui/widgets/memory_observer.cc
```diff
@@ -0,0 +1,222 @@
+/***************************************************************************
+ *   Copyright (C) 2019 PCSX-Redux authors                                 *
```
nicolasnoble, line 2: 2022 :)

src/gui/widgets/memory_observer.cc
```diff
@@ -0,0 +1,222 @@
...
+#include "core/psxmem.h"
+#include "core/system.h"
+
+void PCSX::Widgets::MemoryObserver::draw(const char* title) {
+    if (!ImGui::Begin(title, &m_show)) {
+        ImGui::End();
+        return;
+    }
+
+    uint8_t* memData = g_emulator->m_psxMem->g_psxM;
```
nicolasnoble, line 34: Probably `const uint8_t*` here.

src/gui/widgets/memory_observer.h
```diff
@@ -0,0 +1,75 @@
...
+    static int getMemValue(uint32_t absoluteAddress, const uint8_t* memData, uint32_t memSize, uint32_t memBase, uint8_t stride);
+
+    ScanType m_scanType = ScanType::ExactValue;
+    ScanAlignment m_scanAlignment = ScanAlignment::OneByte;
+    std::vector<AddressValuePair> m_AddressValuePairs;
+    bool m_hex = false;
+    int m_value = 0;
+    bool m_showMemoryEditor = false;
+
+    MemoryEditor m_memoryEditor;
```
nicolasnoble, line 70: We could avoid creating yet another memory editor, and simply use the `JumpToMemory` event to trigger the main memory editor to move, like this:

https://github.com/grumpycoders/pcsx-redux/blob/3b28a1d6be1423ed535d40364aa75dca5d4e0c3c/src/gui/widgets/assembly.cc#L299

src/gui/widgets/memory_observer.cc
```diff
@@ -0,0 +1,222 @@
...
+    uint32_t memSize = 8 * 1024 * 1024;
```
nicolasnoble, line 35: Might be worth doing something like

```c
const uint32_t memSize = 1024 * 1024 * (g_emulator->settings.get<PCSX::Emulator::Setting8MB>() 8 : 2)? 
```

src/gui/widgets/memory_observer.cc
twevs: Ha, I had a variety of project files open as references. :)

twevs:
Thank you for your feedback, I have pushed the changes.

nicolasnoble (dismissed):

PR #736, opened 2022-02-15
Remove overrides for SPU: Streaming, Volume, SPUIRQWait

johnbaumann:
Not sure if these setting changes were leftover from development or if they server some greater purpose. The volume kept defaulting to Loudest which was just a little annoying.


nicolasnoble:
Totally leftovers.

Thanks.

nicolasnoble (approved):

PR #739, opened 2022-02-19
[SIO] Fix memory card block titles

wheremyfoodat:
Before vs after
image
image

Helps the games really discover their identity. No longer do they have to live just going "????????????????????????" when asked for their name.


nicolasnoble:
I've been doing too much Japanese :-D

nicolasnoble (approved):

PR #740, opened 2022-02-19
Added pattern search.

twevs:
Adds a tab to the Memory Observer to search for patterns. If AVX2 instructions are available, provides faster search options for 8- and 16-byte patterns.

### DIFF
--- a/src/gui/widgets/memory_observer.cc
+++ b/src/gui/widgets/memory_observer.cc
@@ -46,7 +46,6 @@ void PCSX::Widgets::MemoryObserver::draw(const char* title) {
             const auto stride = static_cast<uint8_t>(m_scanAlignment);
 
             if (m_addressValuePairs.empty() && ImGui::Button("First scan")) {
-
                 int memValue = 0;
 
                 for (uint32_t i = 0; i < memSize; ++i) {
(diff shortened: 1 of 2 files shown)


twevs:
Sorry for the bother but thinking back, I realized that wanting to check that the comparison of the pattern's and of the contents' shifts yielded all 1s, I unthinkingly (as I was hacking away in an experimental project rather than PCSX-Redux directly) copied it from [here](http://0x80.pl/notesen/2021-02-02-all-bytes-in-reg-are-equal.html#avx2), not realizing that under the 2-clause BSD licence applied to that code, it can be included in a GPL'd project but requires the resulting `all_equal()` function to carry a BSD-compliant notice crediting its author.

(That said, at this point you might want to consider whether it's worth bringing in the AVX2 stuff at all if it'd be a hassle to get working on all builds. The only reason I started out with it was because of the expectation that this would be a very expensive task, but as it turns out, on my machine, the arbitrary-length non-AVX2 search generally takes less than 185 ms for sequences of 8 bytes or more — the longer, the better —, which isn't exactly sloppy.)

wheremyfoodat:
I can try to rewrite the all_equal function so it doesn't contain code from the original if you want

nicolasnoble:
To be fair, we're already doing a bad job crediting all of the dependencies we're currently using. I don't mind taking another dependency that requires crediting, we just need to do our due diligence and add a third party credit tab in the About window.

nicolasnoble (approved):

PR #742, opened 2022-02-21
File API refactoring.

nicolasnoble:


tools/psyq-obj-parser/psyq-obj-parser.cc
```diff
@@ -1159,7 +1161,7 @@ Usage: {} input.obj [input2.obj...] [-h] [-v] [-d] [-n] [-p prefix] [-o output.o
     int ret = 0;
 
     for (auto& input : inputs) {
-        PCSX::File* file = new PCSX::File(input);
+        std::unique_ptr<PCSX::File> file(new PCSX::PosixFile(input));
```
paulsapps, line 1164: `PCSX::PosixFile(input);` should do the trick here, can't see any reason to heap allocate this one

tools/psyq-obj-parser/psyq-obj-parser.cc
```diff
@@ -1185,7 +1187,7 @@ Usage: {} input.obj [input2.obj...] [-h] [-v] [-d] [-n] [-p prefix] [-o output.o
                 }
             }
         }
-        delete file;
+        file.reset();
```
paulsapps, line 1190: if you don't make the above change this line is redundant since the scope of `file` ends here

tools/psyq-obj-parser/psyq-obj-parser.cc
```diff
@@ -531,7 +533,8 @@ std::unique_ptr<PsyqLnkFile> PsyqLnkFile::parse(PCSX::File* file, bool verbose)
     return nullptr;
 }
 
-std::unique_ptr<PsyqLnkFile::Expression> PsyqLnkFile::Expression::parse(PCSX::File* file, bool verbose, int level) {
+std::unique_ptr<PsyqLnkFile::Expression> PsyqLnkFile::Expression::parse(std::unique_ptr<PCSX::File>& file, bool verbose,
```
paulsapps, line 536: I think all of these `file` params should just be `File& file` as `unique_ptr` is forcing a heap allocation for the callers even if they don't need to heap allocate. Also this param seems like it can never be `nullptr` since it is never checked

PR #744, opened 2022-02-22
x64 JIT and interpreter accuracy fixes

wheremyfoodat:
- Almost fully implements load delays for the aligned read instructions and MFC2/CFC2
- More aggressive block invalidation, which works by throwing out all code when the icache is flushed
- Simplicity fixes

Fixes Tekken 2 (all BIOSes), Amidog's tests and the memory card code in OpenBIOS.

Update:
- Also fixes issues related to IRQs on GTE instructions. Fixes visual glitches on several games


src/core/DynaRec_x64/gte_x64.cc
```diff
@@ -196,6 +196,29 @@ static uint32_t MFC2Wrapper(int reg) { return PCSX::g_emulator->m_gte->MFC2(reg)
 
 void DynaRecCPU::recMFC2() {
     if (_Rt_) {
+        const auto loadDelayDependency = getLoadDelayDependencyType(_Rt_);
+        if (loadDelayDependency != LoadDelayDependencyType::NoDependency) {
+            if (_Rd_ != 17) {
+                PCSX::g_system->message("Implement mean GTE behaviour for regs other than SZ1\n");
```
nicolasnoble, line 202: Maybe add the register number / name in the message?

src/core/DynaRec_x64/instructions.cc
```diff
@@ -683,8 +683,67 @@ void DynaRecCPU::recMULTU() {
...
+            call(psxMemRead8Wrapper);
+            break;
+        case 16:
+            call(psxMemRead16Wrapper);
+            break;
+        case 32:
+            call(psxMemRead32Wrapper);
+            break;
+        default:
+            PCSX::g_system->message("Invalid size for memory load in dynarec. Instruction %08x\n", m_psxRegs.code);
```
nicolasnoble, line 706: Shouldn't this be an exception?

src/core/DynaRec_x64/instructions.cc
```diff
@@ -683,8 +683,67 @@ void DynaRecCPU::recMULTU() {
...
+            break;
+    }
+
+    if (_Rt_) {
+        PCSX::g_system->printf("Need delayed load @ %08X\nJIT buffer index: %08X\n", m_pc, gen.getSize());
```
nicolasnoble, line 711: Won't this one spam a lot?

src/core/DynaRec_x64/instructions.cc
```diff
@@ -742,8 +801,18 @@ void DynaRecCPU::recLHU() { recompileLoad<16, false>(); }
 void DynaRecCPU::recLW() { recompileLoad<32, true>(); }
 
 void DynaRecCPU::recLWL() {
-    if (!_Rt_) {
-        error();
+    if (_Rt_ == 0) { // If $rt == 0, just execute the read in case it has side-effects, then return
```
nicolasnoble, line 804: The comment here probably should say "most likely", as the compiler will only generate dummy loads for volatiles.

src/core/DynaRec_x64/recompiler.h
```diff
@@ -181,6 +191,12 @@ class DynaRecCPU final : public PCSX::R3000Acpu {
         }
     }
 
+    virtual void invalidateCache() final {
```
nicolasnoble, line 194: +override

src/core/DynaRec_x64/recompiler.h
```diff
@@ -181,6 +191,12 @@ class DynaRecCPU final : public PCSX::R3000Acpu {
...
+        memset(m_psxRegs.ICache_Addr, 0xff, sizeof(m_psxRegs.ICache_Addr));
+        memset(m_psxRegs.ICache_Code, 0xff, sizeof(m_psxRegs.ICache_Code));
+        m_invalidateBlocks();
```
nicolasnoble, line 197: We may want to also call base, to clear the interpreter's cache.

src/core/DynaRec_x64/recompiler.cc
```diff
@@ -434,4 +533,89 @@ void DynaRecCPU::handleFastboot() {
...
+        NoDep,       NoDep,       NoDep,       NoDep,  // 0x38-0x3F
+    };
+
+    uint8_t dependencyType = NoDep;
+    switch (opcode) {
+        case 0:  // Special instructions
+            dependencyType = specialDependencyList[instruction & 0x3F];
+            break;
+        case 0x10: {  // COP0 instructions also need special handling
+            // We need to emulate the delay if the rs field is 4, ie the instruction is MTC0, and "index" is the
```
nicolasnoble, line 594: "is the" what ? :)

src/core/DynaRec_x64/gte_x64.cc
wheremyfoodat: This is debug logging that's not going to be in the final PR when it's past the draft stage

src/core/DynaRec_x64/instructions.cc
wheremyfoodat: I'll add a static_assert in recompileLoad instead

src/core/DynaRec_x64/instructions.cc
wheremyfoodat: No because it very rarely gets triggered, but it's also debug logging that will be removed in the final PR

src/core/DynaRec_x64/recompiler.h
wheremyfoodat: The 2 memsets should clean the interp cache though?

src/core/DynaRec_x64/instructions.cc
```diff
@@ -683,8 +683,68 @@ void DynaRecCPU::recMULTU() {
...
+            call(psxMemRead32Wrapper);
+            break;
+    }
+
+    if (_Rt_) {
+        m_delayedLoadInfo[m_currentDelayedLoad].active = true;
+
+        switch (size) {
+            case 8:
+                signExtend ? gen.movsx(eax, al) : gen.movzx(eax, al);
```
nicolasnoble, line 712: We probably want `if constexpr (signExtend)` here to make sure.

src/core/DynaRec_x64/recompiler.h
nicolasnoble: Right, that was kind of my point. It's probably better to call the base `R3000Acpu::invalidateCache()` instead of duplicating the memset code, in case the base function changes later on.

nicolasnoble (approved):

PR #758, opened 2022-03-05
Hardware register write size

johnbaumann:
Attempts to correct the write width to register space in accordance with the specifications document.

When accessing the cd player visualizer/sound scope, the shell performs a 16-bit write to DMA PCR. This would cause Redux to discard the upper 16-bits of the register, where as the console would actually perform a 32-bit write.

### DIFF
--- a/src/core/psxhw.cc
+++ b/src/core/psxhw.cc
@@ -555,8 +555,8 @@ void PCSX::HW::psxHwWrite16(uint32_t add, uint32_t rawvalue) {
             } else {
                 psxHu16ref(add) = SWAP_LEu16(value);
                 PSXHW_LOG("*Unknown 16bit write at address %x value %x\n", add, value);
-                return;
             }
+            return;
     }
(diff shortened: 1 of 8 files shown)


src/core/psxhw.cc
```diff
@@ -36,6 +36,24 @@
 
 static inline void setIrq(uint32_t irq) { psxHu32ref(0x1070) |= SWAP_LEu32(irq); }
 
+static constexpr bool between(uint32_t val, uint32_t beg, uint32_t end) {
+    return (beg > end) ? false : (val >= beg && val <= end - 3);
+}
+
+static constexpr bool addressInRegisterSpace(uint32_t address) {
+    uint32_t _add = address & 0x1fffffff;
```
nicolasnoble, line 44: Please don't use identifiers starting with a `_`. They are compiler-reserved.

nicolasnoble (approved):

PR #760, opened 2022-03-05
Add support for Playstation Mouse

wheremyfoodat:
There's at least 1 person somewhere in the globe that wanted this
https://user-images.githubusercontent.com/44909372/156898077-f1c143fc-87e2-44b0-b993-d6834ac9b135.mp4

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,14 +121,19 @@ The codebase still requires a lot of cleanup, and the current product isn't prop
 - memory cards
 - memory card manager
 - XBox controller support
+- digital and analog controller emulation, Playstation Mouse emulation
+- Lua scripting
+- SPU debugger
+- in-app shader editor and built-in crt-lottes shader
+- customizable UI
 - save states
-- save state slots
 
 ### What still requires some work?
 - GLSL GPU
 - proper SPU multithreaded code
 - better customization
 - more generic dynarec
+- hook more emulator functions to Lua
 - ...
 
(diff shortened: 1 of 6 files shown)


nicolasnoble:
This looks so unusable :P Is there a way to try and make it the same coordinates as the viewport?

wheremyfoodat:
> This looks so unusable :P Is there a way to try and make it the same coordinates as the viewport?
Nope, the PS Mouse works by just sending coordinate deltas via SIO. Then each game sets its own mouse area and does stuff.

nicolasnoble:
Maybe having a toggle to capture the mouse completely and shoving the cursor inside the viewport at all times, only keeping the deltas for the mouse input, just like vmware and other virtualbox?

nicolasnoble:
Maybe toggle the capture only when running?

Like, the gui object should only have an internal boolean for `enableRawMouseMotion` / `disableRawMouseMotion`, and then have something like this somewhere:

```c++
    m_listener.listen<Events::ExecutionFlow::Run>([this](const auto& event) { if (m_disableRawMouseMotion) actuallyDisableMouse(); });
    m_listener.listen<Events::ExecutionFlow::Pause>([this](const auto& event) { actuallyEnableMouse(); });
```

nicolasnoble:
Also I don't think it's going to restore properly on config reloading?

nicolasnoble (approved):

PR #788, opened 2022-03-22
CPU meddling and brrrrr

wheremyfoodat:
- Fix GTE flags (Backported from the original MAME implementation)
- Always enable certain exceptions in the interpreter
- Optimize scheduler checks
- Optimize memory accesses
- Optimize function calls in x64 JIT
- Optimize some GTE operations
- More brr coming soon


src/core/psxmem.cc
```diff
@@ -264,183 +262,154 @@ void PCSX::Memory::shutdown() {
...
-        else
-            return PCSX::g_emulator->m_hw->read8(mem);
+    if (pointer != nullptr) {
+        const uint32_t offset = address & 0xffff;
+        return *(pointer + offset);
     } else {
-        p = (char *)(m_readLUT[t]);
-        if (p != NULL) {
-            return *(uint8_t *)(p + (mem & 0xffff));
+        if (page == 0x1f80 || page == 0x9f80 || page == 0xbf80) {
```
nicolasnoble, line 274: Why not masking here? (or even earlier) Wouldn't it be faster than doing the `||` ?

src/core/r3000a.cc
```diff
@@ -313,37 +313,37 @@ void PCSX::R3000Acpu::branchTest() {
...
-        checkAndUpdate(PSXINT_GPUDMA, []() { GPU::gpuInterrupt(); });
-        checkAndUpdate(PSXINT_MDECOUTDMA, []() { g_emulator->m_mdec->mdec1Interrupt(); });
-        checkAndUpdate(PSXINT_SPUDMA, []() { spuInterrupt(); });
-        checkAndUpdate(PSXINT_MDECINDMA, []() { g_emulator->m_mdec->mdec0Interrupt(); });
-        checkAndUpdate(PSXINT_GPUOTCDMA, []() { gpuotcInterrupt(); });
-        checkAndUpdate(PSXINT_CDRDMA, []() { g_emulator->m_cdrom->dmaInterrupt(); });
-        checkAndUpdate(PSXINT_CDRPLAY, []() { g_emulator->m_cdrom->playInterrupt(); });
-        checkAndUpdate(PSXINT_CDRDBUF, []() { g_emulator->m_cdrom->decodedBufferInterrupt(); });
-        checkAndUpdate(PSXINT_CDRLID, []() { g_emulator->m_cdrom->lidSeekInterrupt(); });
+#define checkAndUpdate(irq, act)                                \
```
nicolasnoble, line 316: I'm not convinced the codegen here is better, but, okay.

This sort of lambda should be expanded as-is by the compiler, (especially now that we're on clang everywhere for Release), and have in effect the same result as the macro. If it's not, something else is weird and would have to be investigated.

src/core/sstate.h
```diff
@@ -234,10 +234,9 @@ typedef Protobuf::Message<TYPESTRING("Rcnt"), RcntMode, RcntTarget, RcntRate, Rc
     Rcnt;
 typedef Protobuf::RepeatedField<Rcnt, 4, TYPESTRING("rcnts"), 1> Rcnts;
 typedef Protobuf::Field<Protobuf::UInt32, TYPESTRING("hsync_count"), 2> HSyncCount;
-typedef Protobuf::Field<Protobuf::UInt32, TYPESTRING("spu_sync_count"), 3> SPUSyncCount;
+typedef Protobuf::Field<Protobuf::Int32, TYPESTRING("spu_sync_countdown"), 3> SPUSyncCountdown;
```
nicolasnoble, line 237: Wasn't there another change you had in the pipes that'd bump this? Or is it it?

nicolasnoble (approved):
I am not 100% sure nothing will break from the psxhw changes, but we'll see.

PR #819, opened 2022-04-22
Cdriso bugfix

johnbaumann:
Fix some small bugs I discovered in new cd code.


src/core/cdrom.cc
```diff
@@ -363,7 +363,8 @@ class CDRomImpl : public PCSX::CDRom {
         current = time.toLBA();
 
         for (m_curTrack = 1; m_curTrack < m_iso->getTN(); m_curTrack++) {
-            if (m_iso->getTD(m_curTrack + 1).toLBA() - current >= 150) break;
```
johnbaumann, line 366: This was evaluating as uint32_t, so negative values were greater than 150

nicolasnoble (approved):

PR #827, opened 2022-04-26
Potential fix for over reading CDDA sectors

johnbaumann:
Tinkering with approaches.

### DIFF
--- a/src/cdrom/cdriso.cc
+++ b/src/cdrom/cdriso.cc
@@ -429,6 +429,10 @@ bool PCSX::CDRIso::readTrack(const IEC60908b::MSF time) {
     if (!m_cdHandle || m_cdHandle->failed()) {
         return false;
     }
+    
+    if (this->getTrackType(time) == TrackType::CDDA) {
+        return false;
+    }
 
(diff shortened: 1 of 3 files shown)


src/cdrom/cdriso.cc
```diff
@@ -429,6 +429,10 @@ bool PCSX::CDRIso::readTrack(const IEC60908b::MSF time) {
     if (!m_cdHandle || m_cdHandle->failed()) {
         return false;
     }
+    
+    if (this->getTrackType(time) == TrackType::CDDA) {
```
nicolasnoble, line 433: You should be able to use `m_currentTrack` for this.

nicolasnoble: Oh actually, no, not from this function. I would probably do this in the `readTrack` from cdrom.cc instead.

johnbaumann: Ah, yeah that makes more sense. Couldn't decide the best place between functions to add this 😅

nicolasnoble (approved):

PR #836, opened 2022-05-11
Fixing Appimage generation.

nicolasnoble:


lorn10:
This may be the wrong place but just for the record @nicolasnoble , - since this fixing of the Appimage generation the **build information** is missing. See here regarding current 3929020d:
PCSX-Redux-3929020d-linux_missing_build_info
Otherwise everything is running GREAT in PCSX-Redux under Kubuntu 20.04 LTS. And this is true even for the Windows build 9102 under Wine 7.9. :+1:

nicolasnoble:
Ah, danggit. Thanks for the report.

nicolasnoble:
The version.json file should now be properly generated.

PR #838, opened 2022-05-13
[Pad] Let users enable/disable analog mode manually.

wheremyfoodat:
Required for Gran Turismo 1 and Chrono Cross

### DIFF
--- a/src/core/pad.cc
+++ b/src/core/pad.cc
@@ -173,6 +173,16 @@ void PCSX::Pads::Pad::getButtons() {
         return result ^ 0xffff;  // Controls are inverted, so 0 = pressed
     };
 
+    // Check for analog mode toggle key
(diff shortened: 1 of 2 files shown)


src/core/pad.cc
```diff
@@ -567,6 +567,10 @@ bool PCSX::Pads::Pad::configure() {
         reset();  // Reset pad state when unplugging/replugging pad
     }
 
+    if (m_type == PadType::Analog) {
```
nicolasnoble, line 570: Maybe instead something like:

```c
if (m_type != PadType::Analog) {
  ImGui::PushDisabled();
}

ImGui::Checkbox("Analog mode", &m_analogMode);

if (m_type != PadType::Analog) {
  ImGui::PopDisabled();
}
```

nicolasnoble (approved):

PR #848, opened 2022-05-20
Updated Arch Linux build

KanuX-14:
ImageMagick is needed in order to use the 'convert' command.
Also the '--needed' flag will skip already installed packages.


nicolasnoble:
Hah, thanks!

nicolasnoble (approved):

Issue #546, opened 2021-07-30
Add CHD Support


ducalex:
CHD is a compressed image format that is supported by MAME, Redream, DuckStation, most cdrom-based RetroArch cores, etc. I have [added support for it](https://github.com/ducalex/pcsx-redux/commits/chd-support) using code from [another pcsx fork](https://github.com/libretro/pcsx_rearmed) but I have questions before submitting a pull request for your consideration.

- The code depends on [libchdr](https://github.com/rtissera/libchdr/) which bundles lzma and zlib. Should I add libchdr and lzma to the tree or should it be a remote submodule?
- Where and how should I add a reference to pcsx-rearmed (or the author of the code)?

Thanks

nicolasnoble:
You're more than welcome to make pull requests, yes :-)

The general code style we're using is that most things ought to build out of the box when checking out the repository on development machines. It's acceptable if on Linux and MacOS there's packages being required, as long as they are readily available on common package distributions. On Windows, either nugget packages should be used, or things should compile cleanly from the third_party submodules.

In short, here's sort of the decision tree when adding a new dependency:

- If a nugget package exists, use it on Windows.
- If a commonly, easily installable (apt-get install) Linux package exists use it on Linux.
- If a brew package exists on MacOS, use it on MacOS.
   --> If any of these 3 pre-conditions aren't validated, have a submodule and build from source for the missing cases, using the Makefile at the root, or using a visual studio project file.

I can point to more specific parts of the repository to show more details if required.

With regards to credits and references, I've tried to work it all out in the [AUTHORS](https://github.com/grumpycoders/pcsx-redux/blob/main/AUTHORS) file at the root of the repository, that's shipped with the rest of the codebase. I haven't made it part of the "About" menu yet, but I should.

ducalex:
Perfect, I should be able to figure out the Makefile and VS project changes needed and make sure that it builds cleanly on Windows and Linux but I won't be able to test on MacOS X.

I also see that you are using clang-format so I will make sure it is happy with my changes :).

nicolasnoble:
I don't have MacOS either, I just let the ci yell at me if something breaks. There's a script in the .github folder that installs all the brew dependencies, which hopefully should be enough here.

For the code format, there's a github bot that issues pull requests automatically to fix the style, so it's not a big deal.

lorn10:
@ducalex Just out of curiosity, how can I make such compressed CHD files under Linux? 

[...]

PR #870, opened 2022-06-07
Finally add support for CHD files

RockfordRoe:
It looks like the CHD support thing raised in #546 has never actually had a PR although it has been ready for almost a year, I'm hoping this will be taken care of once and for all.

### DIFF
--- a/src/core/cdriso.cc
+++ b/src/core/cdriso.cc
@@ -1234,6 +1234,81 @@ fail_io:
     return -1;
 }
 
+int PCSX::CDRiso::handlechd(const char *isofile) {
+    uint32_t frame_offset = 0;
+    uint32_t file_offset = 0;
+
+    m_chd_img = (chd_img_t *)calloc(1, sizeof(*m_chd_img));
+    if (m_chd_img == NULL)
+        goto fail_io;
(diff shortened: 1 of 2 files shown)


wheremyfoodat:
Hello. This PR isn't really mergeable. The files edited in the original code don't exist anymore, having long since been refactored. The code also doesn't bring libchd which makes it uncompileable even if merge conflicts are fixed, and it doesn't follow the code style of the rest of the project either.

lorn10:
Perhaps someone should ping @ducalex directly and ask him what the current status is on this interesting matter. :wink:

ducalex:
> Perhaps someone should ping @ducalex directly and ask him what the current status is on this interesting matter. 😉

The code in this PR was definitely not ready to be merged. It worked (at the time) but I couldn't figure out the best way to add libchdr and lzma to the repo (i.e. bundled vs subrepository) and how to properly update the vsprojects to use it.

At this point it's probably best if someone else starts a new PR from scratch!

lorn10:
Thanks for the clarification. So in this case it looks that this PR should be closed.:disappointed:  @wheremyfoodat

PR #883, opened 2022-06-14
Improve assembly screen context menu UX

xyzz:
Changed the menu to use imgui menu widgets:
image
image

Functional changes:
* Copy Address: copies instruction address to clipboard. Personally find this very useful during quick RE iterations
* Go to in Memory View: follows the address in memory view for quick patches, etc
* Toggle Breakpoint: replaces the previous set/remove breakpoint with a single menu item

Breakpoints can also now be toggled by double clicking the address.

### DIFF
--- a/src/gui/widgets/assembly.cc
+++ b/src/gui/widgets/assembly.cc
@@ -652,10 +652,31 @@ void PCSX::Widgets::Assembly::draw(GUI* gui, psxRegisters* registers, Memory* me
                 }
                 ImGui::Text("  %s:%8.8x %c%c%c%c %8.8x: ", section, dispAddr, tc(b[0]), tc(b[1]), tc(b[2]), tc(b[3]),
                             code);
+                auto toggleBP = [&]() mutable {
+                    if (hasBP) {
+                        g_emulator->m_debug->removeBreakpoint(currentBP);
+                    } else {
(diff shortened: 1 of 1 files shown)


src/gui/widgets/assembly.cc
```diff
@@ -652,10 +652,31 @@ void PCSX::Widgets::Assembly::draw(GUI* gui, psxRegisters* registers, Memory* me
                 }
                 ImGui::Text("  %s:%8.8x %c%c%c%c %8.8x: ", section, dispAddr, tc(b[0]), tc(b[1]), tc(b[2]), tc(b[3]),
                             code);
+                auto toggleBP = [&]() mutable {
```
nicolasnoble, line 655: `[&]` or `[=]` are deprecated captures starting in C++20, and we should explicitly capture what we need instead (which, imho, is a bit idiotic, but might as well try and write code that won't break in the next few C++ versions).

Also `mutable` shouldn't be required here? It's only if you're mutating a local capture, like the following:

```c++
std::function<unsigned()> getCounter() {
  unsigned counter = 0;
  return [counter]() mutable -> unsigned { return counter++; };
}
```

nicolasnoble (changes requested):

nicolasnoble (approved):

PR #885, opened 2022-06-14
Add a context menu to editable registers, move PC under GPR

xyzz:
Several things here that I personally find useful during reverse engineering

image

Also move PC under GPR so you don't have switch tabs to access it (it's currently the only other editable register not under GPR).

### DIFF
--- a/src/gui/widgets/registers.cc
+++ b/src/gui/widgets/registers.cc
@@ -59,9 +59,24 @@ void PCSX::Widgets::Registers::draw(PCSX::GUI* gui, PCSX::psxRegisters* register
                     name = PCSX::Disasm::s_disRNameGPR[counter];
                 }
                 counter++;
-                std::string label = fmt::format(f_("Edit##{}"), name);
                 ImGui::Text("%s: %08x", name, reg);
+                std::string contextLabel = fmt::format(f_("Context##{}"), name);
+                if (ImGui::BeginPopupContextItem(contextLabel.c_str())) {
+                    if (ImGui::MenuItem(_("Go to in Assembly"))) {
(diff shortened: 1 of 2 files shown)


src/gui/widgets/registers.cc
```diff
@@ -152,12 +174,6 @@ void PCSX::Widgets::Registers::draw(PCSX::GUI* gui, PCSX::psxRegisters* register
             ImGui::EndTabItem();
         }
         if (ImGui::BeginTabItem(_("Misc"))) {
-            ImGui::Text("pc   : %08x", registers->pc);
```
nicolasnoble, line 155: I wouldn't remove it completely.

src/gui/widgets/registers.cc
```diff
@@ -40,8 +40,10 @@ void PCSX::Widgets::Registers::draw(PCSX::GUI* gui, PCSX::psxRegisters* register
     if (ImGui::BeginTabBar(_("Registers"))) {
         gui->useMonoFont();
         if (ImGui::BeginTabItem("GPR")) {
+            std::vector<std::pair<const char*, uint32_t>> regs;
```
nicolasnoble, line 43: Let's avoid this pattern - it'll be run on every refresh still... instead let's just add a `ImGui::Separator()` at the end of the loop and add `pc` there in a special manner. We should also probably just factor all the "show a register" code into a separate function to be able to reuse it on any register, not just the GPRs.

nicolasnoble (changes requested):

xyzz:
Added separator to PC under GPR, and another copy is also editable under Misc

image

image

nicolasnoble (approved):
Awesome, thanks :)

PR #914, opened 2022-06-27
Tweak VRAM Viewer coordinates.

whiteh0le:
The coordinates displayed in the VRAM viewer are rounded up to the closest number, which means that if you place the cursor more than half a pixel away on either axis, the coordinates of the next pixel will be displayed instead of the coordinates of the current pixel.

# Before
image
*Notice how the cursor is at "0 : 0", but the viewer displays (1 : 1)* 

# After
image
*Notice how the coordinates are now "0 : 0" as expected*


nicolasnoble (approved):

nicolasnoble:
Thanks! :)

PR #919, opened 2022-07-01
Support scratchpad peeking in disassembly view/tracer (Should fix #918)

### LINKED ISSUE #918: CPU trace doesn't output correct scratchpad values
### Describe the bug

When doing CPU traces, contents of the scratchpad memory are outputted as 00s, even when they are not.
For example, it outputs
`80158b34 a7a20010: sh     $v0(000002b0), 0x0010(sp)([1f8003d0] = 0000)`
When it should be outputting
`80158b34 a7a20010: sh     $v0(000002b0), 0x0010(sp)([1f8003d0] = 0036)`
This isn't isolated to just a few examples, it happens on every address that starts with `1f80`. 

### Expected behavior

The tracer should correctly output the contents of the scratchpad memory at that address

### Steps to reproduce the bug

1. Using debug mode and interpreter, run a brief CPU trace
2. Check for values in addresses that start with 1f80
3. Compare them with the actual values in the memory viewer.

### Operating System

Windows 10 Build 19043

### PCSX-Redux version

Version: 9699.20220701.2.x64
Changeset: 4f87c919ac2b82b9173ae3c80715c42c062955c1
Date & time: 2022-07-01 10:58:46

### CPU model

AMD Ryzen 5 4600H

### GPU model

Nvidia GeForce GTX 1650 Ti

### BIOS version

OpenBIOS

### CPU emulation mode

Interpreter

### Additional information

Breath of Fire III (SLES-01304) savestate and CPU trace log file (trace was taken 15 frames after the savestate)
[vram changing.txt](https://github.com/grumpycoders/pcsx-redux/files/9028601/vram.changing.txt)
[SLES01304.zip](https://github.com/grumpycoders/pcsx-redux/files/9028621/SLES01304.zip)

wheremyfoodat:


nicolasnoble:
This shouldn't fix tracing, should it ?

nicolasnoble:
Right so you need the same change in https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/disr3000a.cc#L88-L96. In fact, it's probably better to yeet the function to the base class instead as a helper for all overloads.

nicolasnoble (approved):

PR #920, opened 2022-07-01
Web server: add a reset function to asm symbols

mateusfavarin:
* Change the `assembly/symbols` api to require a `function` variable
* Add a new functionality to reset the symbols loaded in redux


nicolasnoble:
Could you alter `https://github.com/grumpycoders/pcsx-redux/blob/main/tools/ghidra_scripts/ReduxSymbols.java#L29` too? I believe you only need to add `?function=upload` to the URL, right?

mateusfavarin:
> Could you alter `https://github.com/grumpycoders/pcsx-redux/blob/main/tools/ghidra_scripts/ReduxSymbols.java#L29` too? I believe you only need to add `?function=upload` to the URL, right?

Done :)

nicolasnoble (approved):

PR #933, opened 2022-07-21
Fixed delta-over-time searches inverting bytes.

twevs:
The exact restitution of multibyte sequences as they were represented in memory meant that, eg, the short `0001` was being interpreted as 0x0001 rather than 0x100 and therefore not only being inaccurately displayed but also considered lesser than `ff00`, which in turn was interpreted as 0xff00 rather than 0xff.


nicolasnoble:
Derp. Thanks.

This is why we need a lot more unit testing :)

nicolasnoble (approved):

PR #935, opened 2022-07-21
Delta-over-time search: replaced alignment with types.

twevs:
Currently, the user chooses whether to search for 1-, 2- or 4-byte sequences, but lack of typing means that all 1- and 2-byte values are displayed as positive integers.

These changes fix the problem by allowing the user to specify the type they are searching for from among char, uchar, short, ushort and int; signed variants of 1- and 2-byte values are displayed accurately and hex inputs are automatically converted between types.

### DIFF
--- a/src/gui/widgets/memory_observer.cc
+++ b/src/gui/widgets/memory_observer.cc
@@ -184,7 +184,7 @@ void PCSX::Widgets::MemoryObserver::draw(const char* title) {
         }
 
         if (ImGui::BeginTabItem(_("Delta-over-time search"))) {
-            const auto stride = static_cast<uint8_t>(m_scanAlignment);
+            const auto stride = getStrideFromValueType(m_scanValueType);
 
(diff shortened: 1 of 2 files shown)


src/gui/widgets/memory_observer.cc
```diff
@@ -403,6 +405,48 @@ void PCSX::Widgets::MemoryObserver::draw(const char* title) {
...
+            memcpy(&short_val, &memValue, 2);
+            return short_val;
+        case ScanValueType::Ushort:
+            unsigned short ushort_val;
+            memcpy(&ushort_val, &memValue, 2);
+            return ushort_val;
+        case ScanValueType::Int:
+            return memValue;
+        default:
+            assert(false);
```
nicolasnoble, line 445: Please rather a `throw std::runtime_exception` here.

nicolasnoble (approved):

PR #959, opened 2022-08-09
Added Typed Debugger

twevs:


src/gui/widgets/typed_debugger.h
```diff
@@ -0,0 +1,160 @@
...
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+#include "core/debug.h"
+#include "gui/widgets/filedialog.h"
+#include "imgui.h"
+#include "support/eventbus.h"
+
+// Represents the type-name-size tuple for either a struct field or a function argument.
```
nicolasnoble, line 33: Could we nest these structs within the class and / or their own namespace, to avoid globals pollution?

src/gui/widgets/typed_debugger.h
```diff
@@ -0,0 +1,160 @@
...
+    TypedDebugger(bool& show);
+
+    /**
+     * Data importation.
+     */
+
+    Widgets::FileDialog m_importDataTypesFileDialog = {[]() { return _("Import data types"); }};
+    Widgets::FileDialog m_importFunctionsFileDialog = {[]() { return _("Import functions"); }};
+    enum class ImportType { DataTypes, Functions };
+    void import(const char* filename, const ImportType& importType);
```
nicolasnoble, line 95: It might make more sense to take a string_view of the contents of the file, as an argument, instead of the filename, for more flexibility of the import process

Also, taking a reference of an int-like is more expensive than just passing it by value :)

src/gui/widgets/typed_debugger.h
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Data types.
+     */
+
```
nicolasnoble, line 100: Shouldn't most of these be private?

tools/ghidra_scripts/export_to_redux.py
```diff
@@ -0,0 +1,32 @@
+import os
```
nicolasnoble, line 1: Could we add an entry in the README of this folder to quickly describe the function of the script and how to use it?

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,894 @@
...
+        for (size_t i = 0; i < numChildren; ++i) {
+            const auto& field = structs_info[type][i];
+            WatchTreeNode newNode{field.type, field.name, field.size};
+            node->children.push_back(newNode);
+            populate(&node->children.back(), structs_info);
+        }
+    }
+}
+
+bool isInRAM(uint32_t address) {
```
nicolasnoble, line 175: There's a lot of functions there that probably should be declared as `static` to avoid globals pollution.

nicolasnoble (commented):
I've done a first quick pass, will go into more in-depth tomorrow :)

src/gui/widgets/typed_debugger.h
twevs: Oops, I usually do pass types of that size by value when it's semantically appropriate. :)

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+            case ImportType::Functions:
+                m_functions[address] = func;
+                break;
+            default:
+                break;
+        }
+    }
+
+    if (importType == ImportType::DataTypes) {
+        std::sort(m_typeNames.begin(), m_typeNames.end(), [](std::string left, std::string right) {
```
nicolasnoble, line 117: This function signature should be `const std::string & left, const std::string & right`.

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+            for (auto& letter : right) {
+                letter = tolower(letter);
+            }
+            return left.compare(right) < 0;
+        });
+    }
+}
+
+PCSX::Widgets::TypedDebugger::TypedDebugger(bool& show) : m_show(show), m_listener(g_system->m_eventBus) {
+    m_listener.listen<PCSX::Events::ExecutionFlow::Reset>([this](const auto& event) {
```
nicolasnoble, line 130: The `PCSX` namespace here is superfluous. There's a few others in the file.

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+
+    m_listener.listen<PCSX::Events::ExecutionFlow::SaveStateLoaded>([this](const auto& event) {
+        // When loading a savestate, ensure instructions and functions that have been toggled off are disabled, just in
+        // case they weren't when the savestate was created.
+        uint8_t* const memData = g_emulator->m_mem->m_psxM;
+        constexpr uint32_t memBase = 0x80000000;
+
+        for (const auto& disabledInstruction : m_disabledInstructions) {
+            const auto instructionAddress = disabledInstruction.first;
+            constexpr uint8_t nop[4] = {0x00, 0x00, 0x00, 0x00};
```
nicolasnoble, line 154: You can use `mips/common/util/encoder.hh` for these, if you want :) Might be a bit more clear.

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+        auto* newWriteBreakpoint = PCSX::g_emulator->m_debug->addBreakpoint(
+            address, PCSX::Debug::BreakpointType::Write, node->size, _("Write"), logReadsWritesInvoker);
+        m_watchBreakpoints.push_back(newReadBreakpoint);
+        m_watchBreakpoints.push_back(newWriteBreakpoint);
+    }
+
+    if (node->logEntries.size() > 0) {
+        ImGui::TableNextRow();
+        ImGui::TableNextColumn();  // Name.
+        bool open = ImGui::TreeNodeEx(fmt::format(f_("Display log entries##{}{}"), node->name.c_str(), address).c_str(),
```
nicolasnoble, line 280: You're using fmt::format without including its header.

src/gui/widgets/typed_debugger.h
```diff
@@ -0,0 +1,171 @@
...
+        using ArgumentData = std::variant<AddressNodeTuple, RegisterValue>;
+        std::vector<ArgumentData> argData;
+    };
+    std::vector<FunctionBreakpointData> m_displayedFunctionData;
+    std::vector<PCSX::Debug::Breakpoint*> m_functionBreakpoints;
+    std::unordered_map<uint32_t, std::array<uint8_t, 8>> m_disabledFunctions;
+
+    // Returns the name of the function from which the instruction at the given address was emitted if found, an empty
+    // string otherwise.
+    std::string getFunctionNameFromInstructionAddress(uint32_t address);
```
nicolasnoble, line 145: This probably should return an std::string_view instead of an std::string, to avoid allocating and copying strings.

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+    const auto logReadsWritesButtonName = fmt::format(f_("Log reads and writes##{}{}"), node->name.c_str(), address);
+    if (ImGui::Button(logReadsWritesButtonName.c_str())) {
+        PCSX::Debug::BreakpointInvoker logReadsWritesInvoker =
+            [this, node](const PCSX::Debug::Breakpoint* self, uint32_t address, unsigned width, const char* cause) {
+                if (!node) {
+                    return false;
+                }
+
+                const auto pc = g_emulator->m_cpu->m_regs.pc;
+                std::string funcName = getFunctionNameFromInstructionAddress(pc);
```
nicolasnoble, line 251: If changing the return type of this function, don't forget to also change these types.

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+        memcpy(&fieldValue, address, 4);
+        sprintf(s, "%i (0x%2x) \n", fieldValue, fieldValue);
+        ImGui::Text("Value: %s", s);
+    } else if (equals(type, "uint") || equals(type, "u_int") || equals(type, "ulong") || equals(type, "u_long")) {
+        uint32_t fieldValue = 0;
+        memcpy(&fieldValue, address, 4);
+        sprintf(s, "%u (0x%2x) \n", fieldValue, fieldValue);
+        ImGui::Text("Value: %s", s);
+    } else {
+        sprintf(s, "\t> cannot yet print out value of type %s\n", type);
```
nicolasnoble, line 519: Did you rather mean to do ImGui::Text here?

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+    }
+}
+
+void PCSX::Widgets::TypedDebugger::printValue(const char* type, void* address) {
+    static char s[64];
+    if (equals(type, "char")) {
+        int8_t fieldValue = 0;
+        memcpy(&fieldValue, address, 1);
+        sprintf(s, "%d (0x%2x) \n", fieldValue, fieldValue);
+        ImGui::Text("Value: %s", s);
```
nicolasnoble, line 492: Why doing the sprintf to a temporary instead of justusing ImGui::Text's formatting?

src/gui/widgets/typed_debugger.cc
```diff
@@ -0,0 +1,967 @@
...
+
+void PCSX::Widgets::TypedDebugger::draw(const char* title, GUI* gui) {
+    if (!ImGui::Begin(title, &m_show)) {
+        ImGui::End();
+        return;
+    }
+
+    bool showImportDataTypesFileDialog = false;
+    if (m_structs.empty()) {
+        ImGui::TextWrapped(
```
nicolasnoble, line 574: Using C++11's Raw String Literals might look a bit prettier here.

nicolasnoble (dismissed):
I only have a few minor nits, but otherwise this looks really good. Thank you for this excellent new feature. You may merge it at your convenience.

wheremyfoodat (approved):

PR #961, opened 2022-08-09
Kiosk Mode

MeharDT:
Added basic Kiosk Mode (aka "Kids Mode") functionality to the emulator. In its current form, menu access is prevented by disabling hotkey functionality when the setting value is set to true. When this mode is active the emulator is operated via pre-configured user settings and command line flags.

To avoid accidental user changes, the KioskMode setting can only be toggled by editing the pcsx.json file.

### DIFF
--- a/src/core/psxemulator.cc
+++ b/src/core/psxemulator.cc
@@ -115,6 +115,14 @@ int PCSX::Emulator::init() {
 
     m_gpu = settings.get<SettingHardwareRenderer>() ? GPU::getOpenGL() : GPU::getSoft();
 
+    // Enable or disable Kiosk Mode if command line flags are set
+    if (args.get<bool>("kiosk")) {
+        settings.get<SettingKioskMode>() = true;
+    }
+    if (args.get<bool>("no-kiosk")) {
+        settings.get<SettingKioskMode>() = false;
+    }
+
     setPGXPMode(m_config.PGXP_Mode);
     m_pads->init();
     return ret;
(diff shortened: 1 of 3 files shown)


nicolasnoble:
I don't mind the feature at all, despite Redux being targeted for development mainly. It doesn't impede the main features, so that's fine.

It might be relevant to add a command line option to activate / deactivate kiosk mode, using `-kiosk` and `-no-kiosk` to follow the rest of the arguments patterns. It still wouldn't be from the UI to toggle it, but it'd make it a bit less obnoxious to set, instead of going inside the .json file.

One question tho: how exiting works? Just alt-f4 ? Or should it enable a different menu to make it still react on esc, but to have a simpler / more simplified centered menu displayed?

MeharDT:
It's definitely a niche feature. I got some use out of it while tinkering and thought others might as well. :)

> It might be relevant to add a command line option to activate / deactivate kiosk mode, using `-kiosk` and `-no-kiosk` to follow the rest of the arguments patterns. It still wouldn't be from the UI to toggle it, but it'd make it a bit less obnoxious to set, instead of going inside the .json file.

That's a good idea! Should we keep the KioskMode option in the .json file or switch fully to a '-kiosk' command line flag?

> One question tho: how exiting works? Just alt-f4 ? Or should it enable a different menu to make it still react on esc, but to have a simpler / more simplified centered menu displayed?

That's correct, you can still use alt + f4 or alt + tab to exit/move away from the application. If the feature gets use I think it makes sense longterm to build out a custom KioskMode menu with basic entries.

nicolasnoble:
The typical pattern for the arguments is to do something like this:

https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/psxemulator.cc#L109-L116

MeharDT:
> The typical pattern for the arguments is to do something like this:
> 
> https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/psxemulator.cc#L109-L116

Thanks! kiosk and no-kiosk flags have been added. I kept the setting in psxemulator instead of moving it to gui (seems more appropriate in the former in case anyone wants to extend functionality beyond hotkeys/ui).

nicolasnoble (approved):

PR #970, opened 2022-08-11
Give more time for azure pipelines to complete.

nicolasnoble:


nicolasnoble:
Looks like the average is still 30 minutes. The 60+ minutes were probably a glitch.

PR #980, opened 2022-08-16
Typed Debugger: fixed segfault when trying to toggle instructions accessing null pointers

twevs:


src/gui/widgets/typed_debugger.cc
```diff
@@ -330,7 +329,7 @@ void PCSX::Widgets::TypedDebugger::displayBreakpointOptions(WatchTreeNode* node,
                 auto toggleButtonName = functionToggledOff ? fmt::format(f_("Re-enable##{}"), instructionAddress)
                                                            : fmt::format(f_("Disable##{}"), instructionAddress);
                 if (ImGui::Button(toggleButtonName.c_str())) {
-                    auto* instructionMem = memData + instructionAddress - memBase;
+                    auto* instructionMem = g_emulator->m_mem->m_psxM + instructionAddress - 0x80000000;
```
nicolasnoble, line 332: This may be a bit fragile when using the 2/8MB mode.

nicolasnoble (dismissed):

src/gui/widgets/typed_debugger.cc
nicolasnoble: It's more the other way around: in 2MB mode, this won't necessarily do what you want. Some games abuse the fact that the console doesn't know it only has 2MB instead of 8MB, and that the memory warps around at 2MB. Those games will have pointers above the 2MB limit, with the intended effect to have it wrap naturally back to the 2MB space. I specifically had to fix the i-cache code due to this.

See #454

twevs: Ah, okay; I have experienced no problem whatsoever running exclusively in 2 MB mode. `InstructionAddress` here is the value of `pc` when the breakpoint was hit so should it not always be within RAM?

nicolasnoble: So here's what can happen:

- A game loads its binary in the high address space. This means `pc` is within the top 8MB address space.
- This code writes to a memory location where you have a write breakpoint to
- Your breakpoint triggers, but with `pc` located in the high memory space, because that's how it's currently running.

--> this specific computation of the instruction address in memory will be wrong.

The rest of the interpreter and dynarec will properly wrap around in this case, but this code doesn't is my point :)

nicolasnoble:
Ah, the Windows build was broken because you need to merge from HEAD.

nicolasnoble (approved):

PR #983, opened 2022-08-20
GPU cleanse

nicolasnoble:


wheremyfoodat:
I'm on mobile so I can't quote lines of code so bear with me here.

- Certain parts of the code use s16 for vertex coords. This will break certain games like Skullmonkeys which store garbage in the upper bits. Using s11 for both x and y seems to work ok in the GL backend but it might need some testing.
- Current code ignores the case where the start address of a linked list node is unaligned. IIRC the DMA controller force aligns it and there should be a game somewhere in my chat logs that relies on it.
- Missing `if constexpr` instead of plain if when checking if a rect is variable sized
- The `int x, y` pattern can typically be replaced using the `OpenGL::Vector<T, n>` class and its specializations (ivec2, vec3, dvec4, etc). Might be worth adding operator overloads for stuff like vector addition too.

PR #995, opened 2022-08-28
Sio0 rework/pocketstation command additions

johnbaumann:
WIP. Refactored quite a lot of SIO code.

Also experimenting with state machine from OpenMC to attempt expanding memory card command set, to potentially allow for basic pocket station support.


johnbaumann:
Bah, got side tracked. Save state settings need finalized, rest is ok for review.

src/core/memorycard.h
```diff
@@ -0,0 +1,137 @@
...
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.           *
+ ***************************************************************************/
+
+#pragma once
+
+#include <stdbool.h>
```
nicolasnoble, line 22: Not stdbool, no.

src/core/memorycard.h
```diff
@@ -0,0 +1,137 @@
...
+#include <stdint.h>
+
+#include <string>
+
+#include "core/psxemulator.h"
+#include "core/r3000a.h"
```
nicolasnoble, line 28: I don't think any of those 3 headers are necessary here.

src/core/pad.h
```diff
@@ -68,6 +69,13 @@ class Pads {
 
     void setLua(Lua L);
 
+    bool isPadConnected(int pad) {
+        if (pad > m_pads.size())
```
nicolasnoble, line 73: Please fully decorate with {}

src/core/psxemulator.h
```diff
@@ -179,13 +179,16 @@ class Emulator {
     typedef Setting<int, TYPESTRING("MSAA"), 1> SettingMSAA;
     typedef Setting<bool, TYPESTRING("LinearFiltering"), true> SettingLinearFiltering;
     typedef Setting<bool, TYPESTRING("KioskMode"), false> SettingKioskMode;
+    typedef Setting<bool, TYPESTRING("Mcd1Pocketstation"), true> SettingMcd1Pocketstation;
```
nicolasnoble, line 182: Do we want these at true by default? What are the pros and cons?

src/core/sio.h
```diff
@@ -162,29 +72,190 @@ class SIO {
         bool isChained() const { return (allocState & ~1) == 0x52; }
     };
 
-    void getMcdBlockInfo(int mcd, int block, McdBlock &info);
+    static const size_t s_sectorSize = 8 * 16;            // 80h bytes per sector/frame
```
nicolasnoble, line 75: constexpr, not const. Also, please use `c_` as a prefix for constants.

src/core/sio.h
```diff
@@ -162,29 +72,190 @@ class SIO {
...
+        size_t size() { return queue_.size(); }
+
+      private:
+        std::queue<T> queue_;
+    };
+
+    friend MemoryCard;
     friend SaveStates::SaveState SaveStates::constructSaveState();
+
+    static const size_t s_padBufferSize = 0x1010;
```
nicolasnoble, line 228: same here

src/core/sio.h
```diff
@@ -162,29 +72,190 @@ class SIO {
...
+
+    bool isReceiveIRQReady();
+    bool isTransmitReady();
+    inline void scheduleInterrupt(uint32_t eCycle) {
```
nicolasnoble, line 232: +static

src/core/sio.h
```diff
@@ -162,29 +72,190 @@ class SIO {
...
+
+    // Pads
+    uint8_t m_buffer[s_padBufferSize];
+    uint32_t m_bufferIndex;
+    uint32_t m_maxBufferIndex;
+    uint32_t m_padState;
+
+    MemoryCard m_memoryCard[s_cardCount] = {this, this};
+
+    FIFO<8, uint8_t> m_rxFIFO;
```
nicolasnoble, line 258: Purely for cosmetic reasons, I would've swapped the template parameters here.

src/gui/widgets/memcard_manager.h
```diff
@@ -61,11 +61,12 @@ class MemcardManager {
     void getPocketstationIcon(uint32_t* pixels, const SIO::McdBlock& block);
 
     void saveUndoBuffer(std::unique_ptr<uint8_t[]>&& tosave, const std::string& action);
+    void ShowHelpMarker(const char* desc);
 
     std::unique_ptr<uint8_t[]> getLatest() {
-        std::unique_ptr<uint8_t[]> data(new uint8_t[SIO::MCD_SIZE * 2]);
-        std::memcpy(data.get(), g_emulator->m_sio->getMcdData(1), SIO::MCD_SIZE);
-        std::memcpy(data.get() + SIO::MCD_SIZE, g_emulator->m_sio->getMcdData(2), SIO::MCD_SIZE);
+        std::unique_ptr<uint8_t[]> data(new uint8_t[SIO::s_cardSize * 2]);
```
nicolasnoble, line 67: Please use std::make_unique instead. Also this feels like it should be a `Slice` instead, but let's not redesign this :)

src/core/psxemulator.h
johnbaumann: Woops, no. Those should've been defaulted to no, thought I changed these back already.

src/core/sio.h
johnbaumann: Makes sense

src/core/sio.h
johnbaumann: Noted

src/core/memorycard.cc
```diff
@@ -0,0 +1,572 @@
...
+    m_commandTicks++;
+    acknowledge();
+
+    return data_out;
+}
+
+uint8_t PCSX::MemoryCard::tickPS_GetDirIndex(uint8_t value) {
+    uint8_t data_out = Responses::IdleHighZ;
+
+    switch (m_commandTicks) {
```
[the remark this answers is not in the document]
johnbaumann: Yeah not terribly decided. The responses are all basically canned, but at least the framework is here if someone felt the need to add actual logic.

src/core/memorycard.h
```diff
@@ -0,0 +1,149 @@
...
+ *                                                                         *
+ *   You should have received a copy of the GNU General Public License     *
+ *   along with this program; if not, write to the                         *
+ *   Free Software Foundation, Inc.,                                       *
+ *   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.           *
+ ***************************************************************************/
+
+#pragma once
+
+#include "core/psxemulator.h"
```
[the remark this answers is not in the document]
johnbaumann: I give up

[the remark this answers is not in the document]
nicolasnoble: This is an old, weird comment. You've already addressed it.

src/core/pad.h
johnbaumann: Do you decorate only if-else or any if statement in general? Particularly, something like `if (condition) return;`

nicolasnoble: Basically, if it goes on two lines for any reason, then braces should appear. If it's single-line and simple enough, then it's okay not to have braces.

Multi-lines no-braces is a pitfall: https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

johnbaumann: Cool, just seeing where you stand on that. I always try to be pretty explicit with brackets, namespaces, etc., goto fail being a big inspiration - was just pondering how broad to interpret this(and how much code to re-decorate).

nicolasnoble (approved):

PR #1011, opened 2022-10-28
CD-Rom support for psyqo

nicolasnoble:
+ fixing a couple of bugs in Redux.

### DIFF
--- a/src/core/cdrom.cc
+++ b/src/core/cdrom.cc
@@ -1137,7 +1137,7 @@ class CDRomImpl : public PCSX::CDRom {
             memset(m_transfer, 0, PCSX::IEC60908b::DATA_SIZE);
             m_stat = DiskError;
             m_result[0] |= STATUS_ERROR;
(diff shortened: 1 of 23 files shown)


johnbaumann:
Seems to pass all of my usual CDDA games and tests(so far).

Perhaps outside the scope of this PR, but I did notice that attempting to boot invalid disks such as a data cd will cause the bios/shell to hang rather than hit the usual 'please insert a playstation cd-rom' error screen.

johnbaumann (dismissed):

nicolasnoble:
> I did notice that attempting to boot invalid disks such as a data cd will cause the bios/shell to hang rather than hit the usual 'please insert a playstation cd-rom' error screen.

Is this new behavior? This may mean a regression, but then again, I'm not totally sure if I care about this specific case.

nicolasnoble:
Cool! Thanks for checking :)

PR #1023, opened 2022-11-07
Cdrom misc

johnbaumann:
Fixing up some cdrom behavior around command responses. The main focus is currently around disc type detection and to avoid sending the license string for a non-bootable disc, though ideally this would not prevent reading from "unlicensed" discs, i.e. an asset CD when paired with a manually launched binary.


src/core/cdrom.cc
```diff
@@ -909,22 +910,30 @@ class CDRomImpl : public PCSX::CDRom {
                 }
 
                 m_result[0] = m_statP;
-                m_result[1] = 0;
-                m_result[2] = 0;
-                m_result[3] = 0;
+                memset((char *)&m_result[1], 0, 7);
```
nicolasnoble, line 913: Might be easier / safer to do the following:

```c++
memset(m_result, 0, sizeof(m_result);
m_result[0] = m_statP;
```

src/core/cdrom.cc
```diff
@@ -1642,13 +1652,18 @@ void PCSX::CDRom::check() {
...
+            }
+
             break;
         }
     } else {
         IO<File> psxexe(reader.open("PSX.EXE;1"));
         if (!psxexe->failed()) {
             m_cdromId = "SLUS99999";
             exename = "PSX.EXE;1";
+            m_bootable = true;
```
nicolasnoble, line 1666: Maybe the `check` function should display a warning at the end that the disc isn't bootable in case this boolean is still false.

nicolasnoble:
FWIW, asset discs exist on the PS1. They'll usually contain a dummy binary that displays a warning about the disc being a "DLC".

For example, [beatmania APPEND 3rdMIX](https://remywiki.com/CS_bm_3rd).

johnbaumann:
Superseded by #1129 

Leaving branch/PR in-tact until the new code is mostly working.

PR #1049, opened 2022-11-20
Events::ExecutionFlow::Run -> Pause

BatslyAdams:
Tiny fix for the `ExecutionFlow::Pause` Lua event, previous behavior would trigger the `Pause` callback when resuming debugger / `Run`.


nicolasnoble:
Derp! Thanks for the fix.

nicolasnoble (approved):

PR #1095, opened 2022-11-30
Add Simplified Chinese translation file

2601677867:
According to some habits of Chinese programmers, some terms are not translated into Simplified Chinese

Example：Step In , Step Out

### DIFF
--- a/i18n/zh_CN.po
+++ b/i18n/zh_CN.po
@@ -151,7 +151,7 @@ msgstr "无 MSAA"
 
 #: src/core/OpenGL_GPU/gpu_opengl.cc:525
 msgid "{}x MSAA"
-msgstr "{}x MSAA"
+msgstr "{1}x MSAA"
 
 #: src/core/OpenGL_GPU/gpu_opengl.cc:527
 msgid "MSAA"
@@ -393,11 +393,11 @@ msgstr "未绑定"
 
 #: src/core/pad.cc:790
 msgid "Unknown keyboard key {}"
-msgstr "未知键盘按键{}"
+msgstr "未知键盘按键{1}"
 
 #: src/core/pad.cc:795
 msgid "Keyboard {}"
-msgstr "键盘{}"
+msgstr "键盘{1}"
 
 #: src/core/pad.cc:810 src/gui/gui.cc:1627
 msgid "Auto"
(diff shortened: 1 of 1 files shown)


nicolasnoble:
Is "Zhongwén" acceptable to display as the Romanized name? The ō isn't part of the basic list of loaded glyphs in the English UI.

Also, should I load the full, or simplified list of Chinese glyphs?

And which of these fonts should we load? https://fonts.google.com/noto/fonts?noto.region=CN

Thank you for your contribution!

2601677867:
I'm sorry, roman characters shouldn't be in this file, I've fixed that in a new commit

Please use Noto Sans Simplified Chinese font

https://fonts.google.com/noto/specimen/Noto+Sans+SC?noto.region=CN

nicolasnoble:
There's a few minor changes that we'll need to do. The character you've selected for "Circle", ⚪ (unicode character 26AA) isn't available in the font, but this one is available: ◯ (unicode character 25EF) so we probably want to change this.

Also, something that isn't necessarily very clear is the `{}` in strings are used for formatting. So when re-ordering them, you need to add numbers in them to indicate the order change. For example:

```po
msgid "Moved file {}({}) from card {} to card {}"
msgstr "从卡{3} 中移动文件{1}({2})到卡{4}"
```

Finally, the language dropdown menu is using romanization of language names to ensure them being displayed properly regardless of the currently selected font. So I think I'll be using "Zhongwen" as the Hanyu Pinyin writing?

I'll make the proper changes to load the Simplified Chinese noto font.

2601677867:
I changed ⚪ (unicode character 26AA) to "圆圈"
And add numbers in them to indicate the order change

Zhongwen is ok

nicolasnoble:
> I changed white_circle (unicode character 26AA) to "圆圈"

As you wish! Do you want to also have names for the others for consistency? You had symbols too for Cross, Triangle, and Square.

2601677867:
``
#: src/core/pad.cc:815
msgid "Cross"
msgstr "叉叉"

#: src/core/pad.cc:815
msgid "Square"
msgstr "方框"

#: src/core/pad.cc:815
msgid "Triangle"
msgstr "三角"

#: src/core/pad.cc:816
msgid "Circle"
msgstr "圆圈"
``

nicolasnoble (approved):

PR #1113, opened 2022-12-01
Adding Chinese locale support.

nicolasnoble:


nicolasnoble:
It looks like not all glyphs are loaded properly.

image

Do you have any insight of what's missing, @2601677867 ?

The Simplified Chinese ranges being loaded are from https://github.com/ocornut/imgui/blob/bd96f6eac4ad544efb265d7e6bcdb30f99a841c4/imgui_draw.cpp#L2839-L2845

2601677867:
From the picture, it seems that you only added support for Kanji（Japanese HanZi）, because the missing characters in the picture do not exist in the Kanji range（渲，用）

Please try changing line 2845 of the code to

`0x2e80, 0x9FFF, // CJK Ideograms`

sorry for the late reply

nicolasnoble:
It's quite alright, I'll muck around a bit with this :) Thanks!

nicolasnoble:
Okay, there was a few bugs I had to solve, but this is basically good to go now I think.

image


I made a mistake however earlier: the `{}` formatting starts at index 0, not 1. I had to rewrite a bit of the affected file.

PR #1115, opened 2022-12-01
Possible fix for bad window position + size config values

johnbaumann:


nicolasnoble (approved):
We probably want to clean this up further, by properly using the interception of window positions from glfw's callbacks, but this will do for now in terms of stopping the bleeding.

PR #1150, opened 2022-12-11
Fix capture buffer wrap behavior, fix capture buffer left right

johnbaumann:


johnbaumann:
Fixes the bios soundscope and greatly reduces the number of warnings about the capture buffer overflowing.

Some warnings do still occur, typically during hangs/framerate drops(such as resizing the window or saving/loading savestates).

nicolasnoble (approved):

PR #1156, opened 2022-12-14
Lua uvFifoListener ffi

caldog20:
Adds ability to spin up a uvFifoListener (similar to using uvFifo in Lua) but as a server instead of a client. 

- Garbage Collection for the fifo inside the wrapper needs adjusting
- UvFifoListener needs some pending changes to move m_async and tidy up a few things inside the class. This can be done after the current Lua changes are finalized

### DIFF
--- a/src/lua/fileffi.lua
+++ b/src/lua/fileffi.lua
@@ -20,6 +20,7 @@ ffi.cdef [[
 typedef struct { char opaque[?]; } LuaFile;
 typedef struct { uint32_t size; uint8_t data[?]; } LuaBuffer;
 typedef struct { char opaque[?]; } LuaSlice;
+typedef struct { char opaque[?]; } LuaServer;
 
 enum FileOps {
     READ,
@@ -48,6 +49,12 @@ LuaFile* bufferFileEmpty();
 LuaFile* subFile(LuaFile*, uint64_t start, int64_t size);
 
 LuaFile* uvFifo(const char* address, int port);
+
+LuaServer* uvFifoListener();
+void stop(LuaServer* server);
+void start(LuaServer* server, unsigned port, void (*cb)(LuaFile* fifo));
+
+
 
 void closeFile(LuaFile* wrapper);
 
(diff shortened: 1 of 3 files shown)


src/lua/fileffi.lua
```diff
@@ -339,6 +345,40 @@ local function createFileWrapper(wrapper)
...
 end
 
+local function createUvListener(port, cb)
+    if type(port) ~= 'number' then error 'port argument must be an unsigned number' end
+    if (port) == nil then error 'must provide a port' end
+    if cb == nil then error 'must provide a callback function' end
+
+    _callback = function(fifo)  end
+
+    if cb ~= nil then
```
nicolasnoble, line 355: If we check for `cb == nil` before, this check isn't really useful.

Also, we're going to overwrite `_callback` afterward. And it should be declared as `local`.

src/lua/luafile.cc
```diff
@@ -73,6 +74,22 @@ LuaFile* subFile(LuaFile* wrapper, uint64_t start, int64_t size) {
 }
 LuaFile* uvFifo(const char* address, int port) { return new LuaFile(new PCSX::UvFifo(address, port)); }
 
+LuaServer* uvFifoListener() {return new LuaServer(new PCSX::UvFifoListener());}
```
nicolasnoble, line 77: We also need a function to delete a LuaServer. It's not simple as it needs to schedule a stop if the server is running, and then delete in the async close callback.

src/lua/luafile.cc
```diff
@@ -73,6 +74,22 @@ LuaFile* subFile(LuaFile* wrapper, uint64_t start, int64_t size) {
...
+            cb(new LuaFile(fifo));
+        } else {
+            server->m_async.data = server;
+            uv_close(reinterpret_cast<uv_handle_t*>(&server->m_async), [](uv_handle_t* handle) {});
+        }
+    });
+}
+
+void stopListener(LuaServer* server) {
+   server->m_listener->stop();
```
nicolasnoble, line 90: Probably need a status variable to avoid double stops. It needs to be "STARTED", "STOPPING", "STOPPED".

src/lua/fileffi.lua
```diff
@@ -339,6 +345,40 @@ local function createFileWrapper(wrapper)
...
+        _wrapper = wrapper,
+        _proxy = newproxy(),
+        _cb = _callback,
+        _port = port,
+        startListener = function(listener) C.startListener(listener._wrapper, listener._port, listener._cb)  end,
+        stopListener = function(listener) C.stopListener(listener._wrapper) end,
+    }
+    -- Use a proxy instead of doing this on the wrapper directly using ffi.gc, because of a bug in LuaJIT,
+    -- where circular references on finalizers using ffi.gc won't actually collect anything.
+    debug.setmetatable(listener._proxy, { __gc = function()  C.stopListener(listener._wrapper) end })
```
nicolasnoble, line 378: This should (1) call the deleter of the class instead, and (2) it should free the callback in Lua.

PR #1178, opened 2022-12-24
gdb-server.md: add instructions for clion

JohnyTheCarrot:


JohnyTheCarrot:
The reasoning behind the .gdbinit being so weird is because CLion runs the .gdbinit file before it runs `target remote`, so I had to hack it into working. See [this ticket](https://youtrack.jetbrains.com/issue/CPP-7322)

nicolasnoble (approved):

nicolasnoble:
Thanks!

PR #1208, opened 2023-02-05
Improve mac app layout

thankfulmachine:
When downloading an app or disk image from the Web, files get the `com.apple.quarantine` attribute which signals to the OS that the app may be unsafe. The user can still let the app to run from Finder, which removes the attribute and allows future runs to proceed without a prompt.

However, if an app is irregular, the user can't approve it normally and it may get labeled as "damaged". The workaround is to remove the attribute with something like `xattr` in the terminal but that isn't necessary.

Change the `create-app.sh` script to bring the app into a more typical form for macOS and remove files that might interfere with code-signing or needlessly increase the app's size.

- Don't symlink the `pcsx-redux` binary. Make the real binary exist at `./Contents/MacOS/PCSX-Redux`.
  - Remove the now empty `./Contents/Resources/bin`.
- Call the app icon `AppIcon.icns` as is customary.
  - Remove source images after constructing the app icon.
- Move dylibs into `./Contents/Frameworks`.
  - Add this path to the executable's `@rpath`.
  - Use `@rpath` instead of `@executable_path` for library locations.
- Removing the quarantine attribute during the build has no effect.
- Remove the Linux desktop things from `./Contents/Resources/share/applications` as it isn't necessary for macOS.
- Ad-hoc codesign the app after making changes. This must *always* be the very last step.

Update the README about this.

### DIFF
--- a/.github/scripts/create-app.sh
+++ b/.github/scripts/create-app.sh
@@ -1,10 +1,16 @@
 #!/bin/sh
 
-brew install imagemagick dylibbundler
-
 APP=PCSX-Redux
+APPROOT="${APP}.app"
+
+# /usr/libexec needed for PlistBuddy used below.
 PATH="$PATH:/usr/libexec"
 
+# ImageMagik used for converting .ico files to .png files.
+# dylibbundler used for updating load commands for dylib dependencies.
+brew install imagemagick dylibbundler
+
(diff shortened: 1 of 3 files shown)


nicolasnoble (approved):
Thanks!

PR #1240, opened 2023-03-15
Discard oversized writes to SIO data registers

johnbaumann:
Fixes pad issue with Fromage, maybe other things

### DIFF
--- a/src/core/psxhw.cc
+++ b/src/core/psxhw.cc
@@ -457,8 +457,7 @@ void PCSX::HW::write16(uint32_t add, uint32_t rawvalue) {
 
     switch (hwadd) {
         case 0x1f801040:
-            PCSX::g_emulator->m_sio->write8((uint8_t)value);
-            PCSX::g_emulator->m_sio->write8((uint8_t)(value >> 8));
+            PCSX::g_emulator->m_sio->write8((uint8_t)value);  // 8-bit reg, ignore upper 16 bits
             SIO0_LOG("sio write16 %x, %x\n", add & 0xf, value);
             break;
         case 0x1f801044:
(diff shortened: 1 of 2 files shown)


src/core/psxhw.cc
```diff
@@ -478,7 +477,7 @@ void PCSX::HW::write16(uint32_t add, uint32_t rawvalue) {
             SIO0_LOG("sio write16 %x, %x\n", add & 0xf, value);
             break;
         case 0x1f801050:  // rx/tx data register
-            PCSX::g_emulator->m_sio1->writeData16(value);
+            PCSX::g_emulator->m_sio1->writeData8(value); // 8-bit reg, ignore upper 16 bits
```
wheremyfoodat, line 480: Maybe it's best to cast to u8 here like in line 460 for consistency & clarity. One way or another the 2 versions would be equivalent.

src/core/psxhw.cc
```diff
@@ -610,14 +609,11 @@ void PCSX::HW::write32(uint32_t add, uint32_t value) {
...
-            PCSX::g_emulator->m_sio->write8((uint8_t)value);
-            PCSX::g_emulator->m_sio->write8((uint8_t)((value & 0xff) >> 8));
-            PCSX::g_emulator->m_sio->write8((uint8_t)((value & 0xff) >> 16));
-            PCSX::g_emulator->m_sio->write8((uint8_t)((value & 0xff) >> 24));
+            PCSX::g_emulator->m_sio->write8((uint8_t)value);  // 8-bit reg, ignore upper 24 bits
             SIO0_LOG("sio write32 %x\n", value);
             break;
         case 0x1f801050:
-            PCSX::g_emulator->m_sio1->writeData32(value);
+            PCSX::g_emulator->m_sio1->writeData8(value);  // 8-bit reg, ignore upper 24 bits
```
wheremyfoodat, line 616: Same note here

wheremyfoodat (approved):
Looks good to me. Attached 2 comments.
On a relevant note that can be addressed in a separate PR: the way we implement 16 & 32-bit reads from the RX_DATA registers right now is wrong too, though probably doesn't break anything atm. The way a 32-bit read would work for example, is that it would pop a byte from the SIO fifo, and return that + a preview of the next 3 bytes.

wheremyfoodat (approved):

PR #1251, opened 2023-03-24
Memory Editor State Saves

caldog20:


src/gui/gui.cc
```diff
@@ -1268,38 +1263,38 @@ in Configuration->Emulation, restart PCSX-Redux, then try again.)"));
...
+                editor.draw(g_emulator->m_mem->m_wram, 8 * 1024 * 1024);
             }
             counter++;
         }
-        if (m_parallelPortEditor.show) {
+        if (m_parallelPortEditor.m_show) {
             ImGui::SetNextWindowPos(ImVec2(520, 30 + 10 * counter), ImGuiCond_FirstUseEver);
             ImGui::SetNextWindowSize(ImVec2(484, 480), ImGuiCond_FirstUseEver);
-            m_parallelPortEditor.draw(g_emulator->m_mem->m_exp1, 512 * 1024, 0x1f000000);
+            m_parallelPortEditor.draw(g_emulator->m_mem->m_exp1, 64 * 1024);
```
nicolasnoble, line 1276: Probably don't revert this size.

src/gui/gui.h
```diff
@@ -304,24 +335,44 @@ class GUI final {
...
+        MemoryEditorWrapper(GUI *gui, bool &show, size_t &offsetAddr, size_t baseAddr = 0x0000)
+            : m_show(show), m_offsetAddr(offsetAddr), m_baseAddr(baseAddr) {
             editor.OptShowDataPreview = true;
             editor.OptUpperCaseHex = false;
             editor.PushMonoFont = [gui]() { gui->useMonoFont(); };
         }
-        MemoryEditor editor;
+        bool &m_show;
+        size_t &m_offsetAddr;
+        size_t m_baseAddr;
```
nicolasnoble, line 346: Probably should be `const`.

src/gui/gui.cc
caldog20: Oops. This was my mistake during a merge conflict.

src/gui/gui.h
caldog20: Actually you know what, since this needs to be in the wrapper too it might as well also be a reference down to the hex editor class, just a constant one. Is that okay?

nicolasnoble: Well, the advantage of a reference is the ability to change it from somewhere else, but it'll otherwise still cost the dereferencing impact. If the value truly will never change, then we wouldn't gain anything from it being a reference, and will only pay its price.

caldog20: Roger that. I can make the wrapper member const as well as inside the hex editor.

caldog20:
Ready for another review once CI is complete.

src/gui/gui.cc
nicolasnoble: You still need to fix it tho :P

nicolasnoble (approved):

PR #1259, opened 2023-03-28
Update spawn-compiler for better compatibility

xyzz:
- Allow specifying PREFIX env var so you don't need root to install
- Using `./contrib/download_prerequisites` so that the host os doesn't need gmp/mpc/etc installed to build it


nicolasnoble (approved):

nicolasnoble:
Wait, the file `./contrib/download_prerequisites` is missing, right?

xyzz:
It comes from the GCC distribution, usage documented in https://gcc.gnu.org/wiki/InstallingGCC

nicolasnoble:
Ah, right, didn't follow this properly. Makes sense now.

Issue #650, opened 2021-10-10
Spyro 2 PAL does not boot due to ICache emulation code


gameblabla:
Shortly after boot, the game gets stuck on the Insomniac screen :
![spyro2](https://user-images.githubusercontent.com/8717966/136708296-ed56b78e-843d-4374-8125-c99169af68d8.png)

Mednafen used to have this issue in that game because they implemented CdlSync as a Nop and did not return an error.
As a result, the game would lock up like this. (my PR explicitly fixes that https://github.com/grumpycoders/pcsx-redux/pull/648)
However in our case, the issue is different.

It turns out after i merged back your ICache code back to PCSX rearmed (See https://github.com/notaz/pcsx_rearmed/pull/198), that Read_Icache was for some reason causing the game to misbehave (true cause still unknown) and disabling icache emulation would make it work again.

However, because there is no way to disable icache emulation, the game will stay stuck (at least in interpreter mode).

lorn10:
For the record, I can confirm that a self-made (AcetoneISO) **Spyro 2 Gateway to Glimmer** PAL (SCES-02104) ISO image works absolutely fine in PCSX-Redux 5435d5ed and stock SCPH-7502 BIOS v4.1. I have used the needed LibCrypt **sbi** file from [here](https://psxdatacenter.com/sbifiles.html).

So it looks that the original issue is no longer present. If so, this can be closed. @wheremyfoodat 

**Note**, - I think there exist an other issue in recent builds with the PSX OpenBIOS and **Crash Bandicoot Warped**. Under Linux I get in some cases a "DynaRec" error during booting. Or it completely refuses to start the game. The Windows build works under Wine until I exist PCSX-Redux. Then it usually crashes also there.

Spyromain:
I have this issue with the latest version of PCSX-Redux.

The game gets stuck on the Insomniac Games screen and the window is not responding. After a few minutes, the line `Capture buffer is overflowing. Increase CB_SIZE.` appears in the logs and the game continues to run normally but is a bit laggy, I can even start a game until it gets stuck again, thus making the game unplayable.

**PCSX-Redux version:** 14198.20230331.6.x64 (Windows x64)
**Tested BIOS:** openbios and SCPH-7502
**Tested ROM:** SCES-02104, both .cue/.bin and .iso formats
**SBI file included**

**Output logs** (using SCPH-7502 bios + .cue/.bin format + .sbi file):
```
[...]

PR #1263, opened 2023-04-02
Fixing counter overflow.

nicolasnoble:
64 bits numbers obviously won't overflow the way we'd expect them to with 32 bits math.


nicolasnoble:
Hopefully should fix #650.

nicolasnoble:
So this fixes the odd framerate issues in Spyro 2, but not the icache problem in interpreter mode.

PR #1293, opened 2023-04-25
Typed Debugger: jump button for expandable ptrs

Fuuzetsu:
For pointers that were roots of nodes, we would not offer a button to go to them. This is useful to have however, so let's add it.

Indeed, it'd be nice to be able to go to any address that we're rendering, but that's for a potential future refactor...

Before:

xxx

After:

yyy


Fuuzetsu:
Should I do something to fix code coverage or did it fail to merge due to Windows build timing out and we just need to restart that?

wheremyfoodat:
Restarted the Azure build, should merge automatically if the Windows build passes.

Issue #1250, opened 2023-03-23
Doesn't build with -Wformat-security.


Fuuzetsu:
In [nixpkgs](https://github.com/NixOS/nixpkgs), -Wformat-security is on by default. When trying to package this, it failed to build.

```
src/gpu/soft/gpu.cc: In member function 'virtual void PCSX::SoftGPU::impl::debug()':
src/gpu/soft/gpu.cc:947:20: error: format not a string literal and no format arguments [8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wformat-security-Werror=format-security8;;]
  947 |         ImGui::Text(
      |         ~~~~~~~~~~~^
  948 |             _("Debugging features are not supported when using the software renderer yet\nConsider enabling the "
      |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  949 |               "OpenGL "
      |               ~~~~~~~~~
  950 |               "GPU option instead."));
      |               ~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: some warnings being treated as errors
make: *** [Makefile:238: src/gpu/soft/gpu.o] Error 1
```

It'd be nice if it built without having to explicitly disable this hardening feature.

It's probably not really a problem due to the nature of this software but if it's very easy to address, cool.

I tried with 49817037253c318b10b8470fa9e37efa873ba408 but I don't see anything indicating that it got fixed since.

nicolasnoble:
While this is a valid problem (in this case, it should read `ImGui::TextUnformatted`), there's going to be tons of them throughout the codebase...

Fixing all of them will take some time and mechanical effort.

nicolasnoble:
A good way to solve these might be to overload ImGui::Text with the fmt library. I might experiment a bit.

PR #1309, opened 2023-05-21
fix format security warnings

emily-is-my-username:
While packaging pcsx-redux for nixOS, I stumbled across the same problem as described in https://github.com/grumpycoders/pcsx-redux/issues/1250 and fixed it with this quick patch "to make it work on my machine".

It compiles fine with default settings, but it is possible that I didn't find every instance. Also, I am no C++ developer, have no experience with ImGui and just installed pcsx-redux for the first time, so I couldn't check if everything still  looks as intended.

For `ImGui::TextDisabled` and `ImGui::TextWrapped`, I expanded the "shortcuts" explained in [imgui.h](https://github.com/ocornut/imgui/blob/master/imgui.h#LL491C153-L491C153) to create unformatted equivalents, but there may be a better way to do this, feedback is appreciated :)

### DIFF
--- a/src/gpu/soft/gpu.cc
+++ b/src/gpu/soft/gpu.cc
@@ -944,7 +944,7 @@ bool PCSX::SoftGPU::impl::configure() {
 
 void PCSX::SoftGPU::impl::debug() {
     if (ImGui::Begin(_("Soft GPU debugger"), &m_showDebug)) {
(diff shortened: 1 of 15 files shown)


nicolasnoble (approved):
Looks great, thanks!

PR #1315, opened 2023-06-01
Add a memory write path for lua

johnbaumann:


src/core/psxmem.cc
```diff
@@ -410,6 +445,7 @@ void PCSX::Memory::write8(uint32_t address, uint32_t value) {
             }
         } else if ((page & 0x1fff) >= 0x1f00 && (page & 0x1fff) < 0x1f80 && pioConnected) {
             g_emulator->m_pioCart->write8(address, value);
+        } else if (sendWriteToLua(address, 1, value)) {
```
nicolasnoble, line 448: Let's add a `LogClass::HW` log line here to indicate the write has been handled by Lua. Same with the other writes. And maybe the reads too.

src/core/pcsxlua.cc
```diff
@@ -39,6 +39,8 @@ void* getParPtr() { return PCSX::g_emulator->m_mem->m_exp1; }
 void* getRomPtr() { return PCSX::g_emulator->m_mem->m_bios; }
 void* getScratchPtr() { return PCSX::g_emulator->m_mem->m_hard; }
 void* getRegisters() { return &PCSX::g_emulator->m_cpu->m_regs; }
+void* getReadLUT() { return PCSX::g_emulator->m_mem->m_readLUT; }
+void* getWriteLUT() { return PCSX::g_emulator->m_mem->m_writeLUT; }
```
nicolasnoble, line 43: Thinking about it a bit more, these two are problematic because we regenerate them on flushCache: https://github.com/grumpycoders/pcsx-redux/blob/3d27c24b11e90a4d3b821942d291730ddab69b31/src/core/psxmem.cc#L595

We may need a different strategy.

johnbaumann: I'll do some thinking on this, I'm certainly open to suggestions in the meantime.

The main use case I had in mind was for mapping/unmapping memory in the EXP1 region like I do in the pio cart emulation, which should be untouched elsewhere(actually this should probably be called from ``Memory::setLuts`` as well):
https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/pio-cart.cc#L31

Maybe ``Memory::setLuts()`` could also trigger a callback to lua which could allow for overriding/updating when the cache is flushed and when memory is inited?

nicolasnoble: It might be the right way to do it, yes, as `setLuts` isn't called too often. I don't think I have any objection regarding this solution.

src/core/psxmem.cc
johnbaumann: Added logging for reads and writes

src/core/pcsxlua.cc
johnbaumann: Added the lua callback for when ``setLuts`` is called. Will be adding some notes to the documentation as well.

nicolasnoble (approved):

PR #1393, opened 2023-08-12
Can we start using C++23 yet?

nicolasnoble:


wheremyfoodat:
RIP Linux CI.
Is there anything particularly nice in C++23? I like mdspan and expected but the latter is pretty easy to replace.

C++20 has a lot more juice and support still sucks

PR #1400, opened 2023-08-21
Real hardware needs

the-connors:
Display asserts to real console output.


wheremyfoodat (approved):

wheremyfoodat:
Could you make the infinite loop use an asm block while you're at it to make the code mayne compile on Clang eventually?

nicolasnoble:
> Could you make the infinite loop use an asm block while you're at it to make the code maybe compile on Clang eventually?

Eh, it's okay, I can do it separately. It's not good form to ask external contributors to fix unrelated problems.

PR #1401, opened 2023-08-21
Maybe fix MacOS plist

wheremyfoodat:


wheremyfoodat:
I will try to get someone on MacOS 10.15 to test the app bundle

PR #1402, opened 2023-08-22
More cleanups to psyqo::Kernel::abort

nicolasnoble:


wheremyfoodat:
The standard [[noreturn]] attribute can also do the trick

https://en.cppreference.com/w/cpp/language/attributes/noreturn

But __attribute__((noreturn)) should be supported pretty much everywhere that GNU C is supported, so it kind of becomes a styling choice

nicolasnoble:
> The standard [[noreturn]] attribute can also do the trick

That's a good point. I'm just too used to litter the gcc attribute everywhere in OpenBIOS.

Issue #1405, opened 2023-08-24
psyqo PAL not centered.


the-connors:
Using psyqo example code on PAL 5552 machine, screen cut on top, and black on bottom.

Code:
```c++
    psyqo::GPU::Configuration config;
    config.set(psyqo::GPU::Resolution::W320)
        .set(psyqo::GPU::VideoMode::AUTO)
        .set(psyqo::GPU::ColorMode::C15BITS)
        .set(psyqo::GPU::Interlace::PROGRESSIVE);
    gpu().initialize(config);
```

Output:
![image](https://github.com/grumpycoders/pcsx-redux/assets/131186408/2c975ab3-6560-4a0f-b6a9-cbacbddae59a)

https://github.com/ps-iowned/loader-beta-test/issues/2

Report: @danhans42.

nicolasnoble:
Alright, I'm going to need some help with this one. I have a PAL console on which I tested this code, and my capture card doesn't show any problem. Tho I might be able to find some old screen to test with and find the right settings. 

@danhans42? @JonathanDotCel?

danhans42:
@nicolasnoble - no worries - got a few different PAL receivers and setups I can test on. Feel free to throw any tests my way

nicolasnoble:
This was address.

PR #1408, opened 2023-08-24
Add PAL Vertical Offset Fix

danhans42:
Screen output using PAL video was offset off the top of the screen. If statement to choose a different value when PAL is being used (thanks sickle) which addresses this.

Addresses the issue raised here https://github.com/grumpycoders/pcsx-redux/issues/1405


nicolasnoble:
Bit of magical values, but that's okay, we can address this later.

nicolasnoble (approved):

PR #1443, opened 2023-09-12
Fix ghidra exporter script when structure field name is empty

mateusfavarin:


twevs (approved):

tools/ghidra_scripts/export_to_redux.py
```diff
@@ -15,6 +15,8 @@
             type_name = component.getDataType().getName()
             print('type: ' + type_name)
             field_name = component.getFieldName()
+            if field_name is None:
+                field_name = component.getDefaultFieldName()
             print('field name: ' + field_name)
```
nicolasnoble, line 20: The proper python-y syntax would've been

```python
print('field name: ' + field_name if field_name != None else component.getDefaultFieldName())
```

but that's okay :)

PR #1468, opened 2023-11-23
Add support for a shared memory mapping on Kernel & User Memory

Giragast:
This allows me to write to game memory from Unity in realtime, without having to go through GDB, Lua, etc, which are (in my experience) unable to keep up with writes being sent at 50fps+

### DIFF
--- a/src/core/psxmem.cc
+++ b/src/core/psxmem.cc
@@ -89,15 +89,9 @@ int PCSX::Memory::init() {
     m_readLUT = (uint8_t **)calloc(0x10000, sizeof(void *));
     m_writeLUT = (uint8_t **)calloc(0x10000, sizeof(void *));
 
-    uint32_t pid = 0;
-#if defined(_WIN32) || defined(_WIN64)
-    pid = static_cast<uint32_t>(GetCurrentProcessId());
-#elif defined(__linux__) || (defined(__APPLE__) && defined(__MACH__))
-    pid = static_cast<uint32_t>(getpid());
-#endif
-
     // Init all memory as named mappings
-    m_wramShared.init(_("wram"), 0x00800000);
+    bool success = m_wramShared.init("wram", 0x00800000, true);
+    if (success) g_system->message(_("SharedMem failed to share memory for wram, falling back to memory alloc\n"));
     m_wram = m_wramShared.getPtr();
 
     m_exp1 = (uint8_t *)calloc(0x00800000, 1);
--- a/src/core/psxmem.cc
+++ b/src/core/psxmem.cc
@@ -89,7 +89,46 @@ int PCSX::Memory::init() {
     m_readLUT = (uint8_t **)calloc(0x10000, sizeof(void *));
     m_writeLUT = (uint8_t **)calloc(0x10000, sizeof(void *));
 
+#ifdef _WIN32
+    size_t wramSize = 0x00800000;
+    bool doFallbackAlloc = true;
+
+    // Try to create a shared memory mapping, if enabled
+    const bool memoryMappingEnabled = PCSX::g_emulator->settings.get<PCSX::Emulator::SettingSharedMemoryMap>();
+    if (memoryMappingEnabled) {
+        // Create the name of our memory mapping handle which we can search for externally
+        auto handleName = fmt::format("pcsx-redux-{}", GetCurrentProcessId());
+        // Create the memory mapping handle
+        m_wramFileHandle = CreateFileMappingA(INVALID_HANDLE_VALUE, nullptr, PAGE_READWRITE, 0,
+            static_cast<uint32_t>(wramSize), handleName.c_str());
+        if (m_wramFileHandle != INVALID_HANDLE_VALUE) {
+            // Create a view of the memory mapping
+            void *base_pointer = MapViewOfFileEx(m_wramFileHandle, FILE_MAP_READ | FILE_MAP_WRITE, 0, 0,
+                static_cast<uint32_t>(wramSize), nullptr);
+            // Validate success and assign the view to m_wram
+            if (base_pointer != nullptr) {
+                doFallbackAlloc = false;
+                m_wram = (uint8_t *)base_pointer;
+            } else {
+                CloseHandle(m_wramFileHandle);
+                m_wramFileHandle = nullptr;
+                g_system->message("MapViewOfFileEx failed, falling back to memory alloc, last error: %d\n",
+                    (int)GetLastError());
+            }
+        } else {
+            m_wramFileHandle = nullptr;
+            g_system->message("CreateFileMappingA failed, falling back to memory alloc, last error: %d\n",
+                (int)GetLastError());
+        }
+    }
+
+    if (doFallbackAlloc) {
+        m_wram = (uint8_t *)calloc(wramSize, 1);
+    }
+#else
     m_wram = (uint8_t *)calloc(0x00800000, 1);
+#endif
+
     m_exp1 = (uint8_t *)calloc(0x00800000, 1);
     m_hard = (uint8_t *)calloc(0x00010000, 1);
     m_bios = (uint8_t *)calloc(0x00080000, 1);
@@ -239,7 +278,21 @@ The distributed OpenBIOS.bin file can be an appropriate BIOS replacement.
 }
 
 void PCSX::Memory::shutdown() {
+#ifdef _WIN32
+    if (m_wramFileHandle != nullptr) {
+        bool result = static_cast<bool>(UnmapViewOfFile(m_wram));
+        if (!result) {
+            g_system->printf("Failed to unmap view of Kernel & User Memory.\n");
+        }
+        m_wram = nullptr;
+        CloseHandle(m_wramFileHandle);
+        m_wramFileHandle = nullptr;
+    } else {
+        free(m_wram);
+    }
+#else
     free(m_wram);
+#endif
     free(m_exp1);
     free(m_hard);
     free(m_bios);
--- a/src/core/psxmem.cc
+++ b/src/core/psxmem.cc
@@ -97,13 +97,14 @@ int PCSX::Memory::init() {
 #endif
 
     // Init all memory as named mappings
-    m_wram.init(fmt::format("pcsx-redux-wram-{}", pid).c_str(), 0x00800000);
+    m_wramShared.init(_("wram"), 0x00800000);
+    m_wram = m_wramShared.getPtr();
 
     m_exp1 = (uint8_t *)calloc(0x00800000, 1);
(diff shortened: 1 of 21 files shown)


nicolasnoble:
So, several comments here.

- I'm not at all opposed to this idea. In fact, I believe it's a great addition. This was low on my TODO list, but I definitely had it in mind. But this means I'd be a bit more opiniated on how to do it :-)
- We may simply want to do it all the time. There's little downside in having the shared memory on, as opposed to just using malloc. I'd just drop the setting to enable it, and simplify the codepath in general.
- We may still want to do this in a cross-platform way, not just for Windows. It feels like we should have a new class in the `src/support` directory, which would allow to RAII the shared memory pages.
  - All allocated memory, not just the main RAM, probably want to have its own shared memory object. The name pattern could reflect that. We could even allow the SoftGPU implementation to have its VRAM shared too.
  - The Unix version of shared memory pages is using the `shm_open` API call. I can help writing this if needed.

Suggestions:

- Yeet common code to `src/support/sharedmem.{h,cc}` (please notice the different license scheme in this directory). The `SharedMem` class could simply take the filename pattern to use, and the size of the allocation, and expose the memory pointer.
- Create `src/support/sharedmem-windows.cc` for the Windows-specific implementation of the methods of the class. Avoid splitting functions with `#ifdef`, and instead have the whole file being disabled with a single `#ifdef` at its top.
- Change `psxmem.{h,cc}` to use this new `SharedMem` object, allowing for RAII to destroy it cleanly.
- If you're not comfortable writing `src/support/sharedmem-unix.cc` using `shm_open` / `mmap` / `mmunmap` / `shm_unlink`, I can take care of this part. The actual filename prefix for Unix needs to be `std::filesystem::temp_directory_path()`.

Giragast:
Okay, I've refactored things into something that's hopefully sensible and there's a few questions I have left

+ I'm assuming the implementation of linux is the same as macosx (from what I've read) - do you want `sharedmem-macosx.cc` (which is the old base behaviour currently) removed and rename `sharedmem-linux.cc` to `sharedmem-nix.cc` or `sharedmem-other.cc` or similar?
+ `g_emulator->m_mem->m_wram.m_mem` has clunky naming, so you want `SharedMem::m_mem` renamed to something else?
+ I've implemented `m_wram` with `SharedMem` and I'm torn between two ways to reference the memory:

```
//HEADER
  public:
    SharedMem m_wram;  // Kernel & User Memory (8 Meg)
//SOURCE
  //init
  m_wram.init(fmt::format("pcsx-redux-wram-{}", pid).c_str(), 0x00800000);
  //referenced as
  g_emulator->m_mem->m_wram.m_mem
  //no clean-up needed
```

VS

```
//HEADER
  public:
    SharedMem m_wramShared;  // Kernel & User Memory (8 Meg)
    uint8_t *m_wram = nullptr;
//SOURCE
  //init
  m_wramShared.init(fmt::format("pcsx-redux-wram-{}", pid).c_str(), 0x00800000);
  m_wram = m_wramShared.m_mem;
  //referenced as
  g_emulator->m_mem->m_wram
  //clean-up
  m_wram = nullptr;
```
+ Is the naming format `pcsx-redux-[memtype]-[pid]` okay?
+ Once everything is good, do you have a list of memory blocks to convert to `SharedMem`? I'm assuming `m_wram`, `m_exp1`, `m_bios` and `m_hard`. If you want `m_writeLUT` and `m_readLUT` then `SharedMem` needs to be templated, which should be fine

nicolasnoble:
> Okay, I've refactored things into something that's hopefully sensible and there's a few questions I have left
> 
> * I'm assuming the implementation of linux is the same as macosx (from what I've read) - do you want `sharedmem-macosx.cc` (which is the old base behaviour currently) removed and rename `sharedmem-linux.cc` to `sharedmem-nix.cc` or `sharedmem-other.cc` or similar?

We could go with sharedmem-unix.cc, and have the guard for it be "not windows". We might want to split it off further at some point in the future, because Linux now has a "better" way to do this using a new API I've never used before, but that's not really important for now.

> * `g_emulator->m_mem->m_wram.m_mem` has clunky naming, so you want `SharedMem::m_mem` renamed to something else?

I'm fine with this for now. I'm thinking maybe we can try and add a casting operator on the SharedMem object and just "use" it as a pointer directly, but I'd need to experiment with this.

> * I've implemented `m_wram` with `SharedMem` and I'm torn between two ways to reference the memory:
> 
> ```
> //HEADER
>   public:
>     SharedMem m_wram;  // Kernel & User Memory (8 Meg)
> //SOURCE
>   //init
>   m_wram.init(fmt::format("pcsx-redux-wram-{}", pid).c_str(), 0x00800000);
>   //referenced as
>   g_emulator->m_mem->m_wram.m_mem
>   //no clean-up needed
> ```
> 
> VS
> 
> ```
> //HEADER
>   public:
>     SharedMem m_wramShared;  // Kernel & User Memory (8 Meg)
>     uint8_t *m_wram = nullptr;
> //SOURCE
>   //init
>   m_wramShared.init(fmt::format("pcsx-redux-wram-{}", pid).c_str(), 0x00800000);
>   m_wram = m_wramShared.m_mem;
>   //referenced as
>   g_emulator->m_mem->m_wram
>   //clean-up
>   m_wram = nullptr;
> ```

First, I'm wondering if we shouldn't just have the `-[pid]` suffix be added by the SharedMem implementation, making it a mandatory suffix that is.

Then I think we should go with your second option for now. As per my comment prior this one, this may just be temporary if we manage to pull off a casting operation that'd work everywhere transparently, which would make this problem moot.

> * Is the naming format `pcsx-redux-[memtype]-[pid]` okay?
Yup!

> * Once everything is good, do you have a list of memory blocks to convert to `SharedMem`? I'm assuming `m_wram`, `m_exp1`, `m_bios` and `m_hard`. If you want `m_writeLUT` and `m_readLUT` then `SharedMem` needs to be templated, which should be fine

Let's make this PR just convert `m_wram` first, in order to work incrementally. Then we can see about migrating the rest, once we iron out the kinks. about things like maybe casting. Templating it can come after.

Giragast:
Should be ready for another pass now 🤞

src/core/psxmem.cc
```diff
@@ -89,7 +89,17 @@ int PCSX::Memory::init() {
     m_readLUT = (uint8_t **)calloc(0x10000, sizeof(void *));
     m_writeLUT = (uint8_t **)calloc(0x10000, sizeof(void *));
 
-    m_wram = (uint8_t *)calloc(0x00800000, 1);
+    uint32_t pid = 0;
```
nicolasnoble, line 92: Looks like this is no longer needed.

src/core/psxmem.h
```diff
@@ -188,6 +189,9 @@ class Memory {
 
     // hopefully this should become private eventually, with only certain classes having direct access.
   public:
+    // Shared memory wrappers, pointers below point to these where appropriate
+    SharedMem m_wramShared;
```
nicolasnoble, line 193: Probably keep it private for now.

src/support/sharedmem.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace PCSX {
+
+class SharedMem {
+  public:
+    SharedMem() {}
+    ~SharedMem();
+
+    void init(const char* id, size_t size);
+    std::string getSharedName(const char* id, uint32_t pid);
```
nicolasnoble, line 40: This is probably private.

src/support/sharedmem-unix.cc
```diff
@@ -0,0 +1,89 @@
...
+            // fd is ready, reserve the memory we need
+            int result = ftruncate(m_fd, static_cast<off_t>(size));
+            if (result < 0) {
+                shm_unlink(m_sharedName.c_str());
+                close(m_fd);
+                m_fd = -1;
+                g_system->message("ftruncate failed, falling back to memory alloc, size: %zu\n", m_size);
+            } else {
+                // ftruncate completed, now map the memory at 0 offset
+                void* basePointer = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, m_fd, 0);
```
nicolasnoble, line 59: `nullptr`

src/support/sharedmem-unix.cc
```diff
@@ -0,0 +1,89 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+*/
+
+#if !defined(_WIN32) && !defined(_WIN64)
+
+#include "support/sharedmem.h"
+#include "core/system.h"
```
nicolasnoble, line 30: The problem in bringing `core/system.h` here is it'd create a cyclic dependency between library projects. We should just not warn if we need to fallback; there's quite nothing the user might be able to do. At *best* we could imagine having a boolean saying the shared memory worked, and report it up in the UI if we really wanted to.

src/support/sharedmem-windows.cc
```diff
@@ -0,0 +1,84 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+*/
+#if defined(_WIN32) || defined(_WIN64)
+
+#include "support/sharedmem.h"
+#include "support/windowswrapper.h"
+#include "core/system.h"
```
nicolasnoble, line 30: Same comment as with `-unix.cc`, we can't use `core/system.h` safely from `support/*`.

src/core/psxmem.cc
```diff
@@ -89,7 +89,17 @@ int PCSX::Memory::init() {
...
+#if defined(_WIN32) || defined(_WIN64)
+    pid = static_cast<uint32_t>(GetCurrentProcessId());
+#elif defined(__linux__) || (defined(__APPLE__) && defined(__MACH__))
+    pid = static_cast<uint32_t>(getpid());
+#endif
+
+    // Init all memory as named mappings
+    m_wramShared.init(_("wram"), 0x00800000);
```
nicolasnoble, line 100: No, let's avoid making it "wram" a localized string. This would break external scripts depending on the selected locale.

nicolasnoble:
Just one last kind of medium issue: `calloc` will inherently ensure the allocated memory is initialized to zero (`memset` that is), which I don't think the shared memory code currently does.

nicolasnoble:
The rest is just minor nits, looks good otherwise :)

nicolasnoble (approved):

PR #1506, opened 2023-12-25
Add optional labels to describe breakpoints

descawed:
Per #1502 . My C++ is pretty rusty, so let me know if anything looks weird. In particular, I don't like duplicating `addBreakpoint`, but if I put the new parameter at the end, breakpoints.cc has to duplicate the default invoker, and if I put it before invoker without the overload, all the other usages of the function have to change.

As for the feature itself, I put the label to the right of the buttons instead of adding it to the name. I feel like, if the window isn't wide enough to show everything, it's better for it to start by cutting off the end of the label than the Remove and Disable buttons. I also prefer to keep the buttons of each row aligned roughly in the same columns.

### DIFF
--- a/src/gui/widgets/breakpoints.cc
+++ b/src/gui/widgets/breakpoints.cc
@@ -19,11 +19,26 @@
 
 #include "gui/widgets/breakpoints.h"
 
-#include "core/debug.h"
+#include "fmt/format.h"
 #include "imgui.h"
 #include "support/imgui-helpers.h"
 
 static ImVec4 s_currentColor = ImColor(0xff, 0xeb, 0x3b);
+
+void PCSX::Widgets::Breakpoints::showEditLabelPopup(const Debug::Breakpoint* bp, int counter) {
+    std::string name = bp->name();
+    std::string title = fmt::format(f_("Edit label of breakpoint {}##{}"), name, counter);
+    if (ImGui::BeginPopupModal(title.c_str(), nullptr, ImGuiWindowFlags_AlwaysAutoResize)) {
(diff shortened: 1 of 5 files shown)


src/gui/widgets/breakpoints.cc
```diff
@@ -94,6 +94,13 @@ void PCSX::Widgets::Breakpoints::draw(const char* title) {
             buttonLabel = _("Enable##") + name;
             if (ImGui::Button(buttonLabel.c_str())) bp->enable();
         }
+        ImGui::SameLine();
+        const std::string& label = bp->label();
+        if (bp->enabled()) {
+            ImGui::Text(" %s", label.c_str());
```
nicolasnoble, line 100: Might be worth making the label editable somehow? Like, if the user clicks on it, open a popup to change it?

There's some code sample for this in the registers.cc code, when editing the values of registers.

descawed: Edit functionality added.

descawed:
Aside from the edit functionality, I made a couple other changes:

- Pressing enter in the Address or Label fields will add the breakpoint. This is a small QoL change that I thought would be nice.
- There was a bug where the breakpoint buttons didn't work correctly if multiple breakpoints have the same name. This has actually been an issue all along with the Remove and Disable buttons, not just with these changes. My first thought was to prevent adding multiple breakpoints with the same address, width, and type. However, I thought maybe that was useful if e.g. different Lua scripts wanted to take action when the same instruction is executed (although, if there are multiple breakpoints on the same address, does the debugger trigger them all?). I ended up just adding a counter to make the widget IDs unique.

nicolasnoble (approved):

PR #1559, opened 2024-01-28
More lua additions

johnbaumann:


nicolasnoble:
So, I'm in a bit of a pickle here. The cdrom refactoring branch has a change there to make the cycle counter 64 bits. Changing this later would be an API breaking change, but I'm still not ready to merge the cd refactoring one. Might be worth doing a last push to finish it.

johnbaumann:
> So, I'm in a bit of a pickle here. The cdrom refactoring branch has a change there to make the cycle counter 64 bits. Changing this later would be an API breaking change, but I'm still not ready to merge the cd refactoring one. Might be worth doing a last push to finish it.

Would it simplify things if I were to do either/both of the following: Pre-empt the change to the cycle counter and simply change the return type of `getCPUCycles()` to `uint64_t`, or, assuming the change is relatively small, piggy back the 64-bit cycle counter change into this PR?

nicolasnoble:
So the problem with changing the return type is that the behavior will still be different after the switch: right now, with a 32 bits counter, it'll roll over every ~2 minutes, whereas with a 64 bits counter, it'd never roll over (or rather, will take a few millennias).

And the problem with cherry-picking the change from the cd-rom branch is that the it's a savestate breaking change, and that'd be too many breaking changes in a row, which I'd like to avoid.

johnbaumann:
> So the problem with changing the return type is that the behavior will still be different after the switch: right now, with a 32 bits counter, it'll roll over every ~2 minutes, whereas with a 64 bits counter, it'd never roll over (or rather, will take a few millennias).
> 
> And the problem with cherry-picking the change from the cd-rom branch is that the it's a savestate breaking change, and that'd be too many breaking changes in a row, which I'd like to avoid.

I'm ok with putting this pr on ice.

johnbaumann:
CPU cycles now returns a 64-bit value, tested working locally.

nicolasnoble (approved):

PR #1572, opened 2024-02-21
More enums, some tidy on registers.cc, dr. hell adsr

johnbaumann:
Draft pr, still need to finalize and clean some things


src/spu/adsr.h
johnbaumann, line 1: Might make more sense to just generate these at runtime

nicolasnoble: Well, they need to be `constexpr` at the least. But compile-time tables like that are fine.

nicolasnoble (approved):

PR #1611, opened 2024-03-10
Multitap driver

johnbaumann:
Initial efforts. Currently lacks config mode. Draft request for picking at and discussing changes required.


src/mips/psyqo/advancedpad.hh
```diff
@@ -0,0 +1,193 @@
...
+    constexpr uint8_t output_default(uint8_t ticks);
+    void processChanges(Pad pad);
+    void readPad();
+    uint8_t transceive(uint8_t data_out);
+    bool waitForAck();  // true if ack received, false if timeout
+
+    // bool m_configMode[8] = {false, false, false, false, false, false, false, false};
+    uint16_t m_padData[8][4];
+    eastl::function<void(Event)> m_callback;
+    bool m_connected[8] = {false, false};
```
johnbaumann, line 187: Forgot to init the other 6 elements

src/mips/psyqo/src/advancedpad.cpp
```diff
@@ -0,0 +1,305 @@
...
+void psyqo::AdvancedPad::readPad() {
+    uint8_t data_in, data_out;
+    uint8_t port_dev_type[2] = {PadType::None, PadType::None};
+
+    syscall_memset(m_padData, 0xff, sizeof(m_padData));
+
+    uint8_t *pad_data;
+    static constexpr unsigned pad_data_width = 8;
+
+    for (uint16_t port = 0; port < 2; port++) {
```
nicolasnoble, line 176: It looks like this scans all ports? Which is different from what the .hh says, about alternating ports per frame. Also, `unsigned`.

src/mips/psyqo/src/advancedpad.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    SIO::Ctrl = SIO::Control::CTRL_IR;
+    SIO::Baud = 0x88;  // 250kHz
+    SIO::Mode = 0xd;   // MUL1, 8bit, no parity, normal polarity
+    SIO::Ctrl = 0;
+
+    using namespace timer_literals;
+
+    Kernel::Internal::addOnFrame([this]() {
+        readPad();
```
nicolasnoble, line 55: So one detail here is that the callbacks put on `addOnFrame` are run at the beginning of every drawn frame. If the software runs at 10FPS, then it'll be called 10 times per second, which may make it lose events, potentially. Same as with the SimplePad, really. But we can change this later.

src/mips/psyqo/src/advancedpad.cpp
johnbaumann: Huh, guessing the description in the header is a mix of leftover from simplepad and/or copilot hallucination, mixed with my own derp. Judging from simplepad's description, this is mentioned as a drawback, so I'll just nuke the comment about alternating frames unless this seems like desirable behavior. simplepad's description below:

 * @details This class is meant to be used as a singleton, probably in
 * the `Application` derived class. It is a simple thunk to the BIOS'
 * PAD interface, and has the same caveats, namely that it should be
 * initialized prior using the BIOS' memory card functions, and that
 * polling will alternate between the two pads at each frame when two
 * pads are connected, which can introduce input lags.

src/mips/psyqo/examples/multitap/multitap.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // Our pad reader.
+    psyqo::AdvancedPad m_input;
+};
+
+// We only have a single since to display the status of the pads.
+class PadTestScene final : public psyqo::Scene {
+    // Since there's only a single scene, we won't need to override the `start`
+    // or `teardown` methods. We will do all the initialization in the application.
+    int framecount = 1;
+    int padindex = 0;
```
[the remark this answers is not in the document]
johnbaumann: Later on in the code I iterate through the pads every ~5 seconds to find the next connected device to display data per that pad, resting on 0 if no pad is found. Not really sure of a more proper C++ way to do this, or if it even makes sense to do in the first place

```C++
    if (framecount++ % (60 * 5) == 0 ||
        !padTest.m_input.isPadConnected(static_cast<psyqo::AdvancedPad::Pad>(padindex))) {
        // Iterate through the pads, stop on next connected pad
        while (!(padTest.m_input.isPadConnected(static_cast<psyqo::AdvancedPad::Pad>(++padindex))) && padindex < 8)
            ;
        if (padindex > 7) {
            padindex = 0;
        }
    }
```

src/mips/psyqo/advancedpad.hh
```diff
@@ -0,0 +1,193 @@
...
+ * @details This class is meant to be used as a singleton, probably in
+ * the `Application` derived class. It does not use the BIOS'
+ * PAD interface. Instead, it uses the SIO interface directly, and
+ * can therefore support more device types including multitaps. Polling
+ * alternates between the two pads/ports on each frame.
+ */
+
+class AdvancedPad {
+  public:
+    enum Pad { Pad1a, Pad1b, Pad1c, Pad1d, Pad2a, Pad2b, Pad2c, Pad2d };
```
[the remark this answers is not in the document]
johnbaumann: Ah, you've answered my question here, really.

nicolasnoble (approved):
Just two little nits, but otherwise it looks fine.

src/mips/psyqo/advancedpad.hh
```diff
@@ -0,0 +1,211 @@
...
+
+    enum class TapReadMode {
+        Passthrough,
+        BufferedTransferAll,
+    };
+
+    /**
+     * @brief Initializes the pads.
+     *
+     * @details This will initialize the pads polling by calling the BIOS'
```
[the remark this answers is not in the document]
johnbaumann: Revised, but possibly lacking in detail.

PR #1633, opened 2024-03-27
makefile: include .c sources in DEPS

emily-is-my-username:
Assigning instead of appending seems unintentional here and no `.dep` files are created for `.c` source files when running `make dep` because of it .

This has not caused any problems for me, it just caught my eye.

I'm sorry for the barrage of micro-PRs, this is the last one :smile:


nicolasnoble:
Micro PRs are great, no worries at all, thanks for the fixes :)

nicolasnoble (approved):

PR #1637, opened 2024-03-31
allow version info file without updateInfo

emily-is-my-username:
Currently, parsing the `version.json` fails if there is no `updateInfo` key.
When packaging for nix, I want to disable update checks but still show some version information in the "About" window.
Putting broken URLs in there feels too hacky for my taste, and it looks easy enough to differentiate between "failed parsing" and "successful parsing without the update info".

I'm not sure if this counts as trivial, please let me know if I should open an issue for discussion instead.


nicolasnoble:
I'm fine with that. The current code is a bit too naive anyway. I'm in the process of ditching AppCenter fwiw, so I'll have to poke at this in the near future.

nicolasnoble (approved):

PR #1641, opened 2024-04-12
Add homebrew formula for GDB with Mips support

DanEEStar:
On a Mac with a ARM-Cpu trying to install GDB with `brew install gdb` will fail with: "gdb: The x86_64 architecture is required for this software."

But by compiling GDB with Mips support similar to the other Homebrew formula it works.

With this self-compiled GDB I could successfully step-debug the "Psy-Q Cube" example project from the VS Code template.


nicolasnoble:
I don't mind merging this in, but it feels like this is a bug to report upstream to homebrew.

nicolasnoble (approved):

nicolasnoble (approved):

PR #1642, opened 2024-04-13
makefile: Add absolute path for bios link scripts

croyzor:
I'm not sure what the difference is between my laptop and the CI, but locally this was resulting in build errors saying the link scripts can't be found


nicolasnoble:
Yeah that's a good change whatever happens.

nicolasnoble (approved):

PR #1659, opened 2024-06-04
OpenBIOS: Add System 573 support

spicyjpeg:
This PR adds several compile-time options to OpenBIOS in order to make it compatible with Konami's System 573 arcade board, which requires slightly different RAM initialization as well as some additional code to periodically clear the system's watchdog during startup.

I have added the following options:

- `BOARD=system573`: switches out the boot stub for one that properly initializes the 573's system bus, adds watchdog code and removes PIO booting support. The watchdog is cleared in roughly the same places as the original 573 kernel.
- `BOOT_MODE=fast`: alias for `FASTBOOT=true`, removes the shell but not the CD-ROM booting code.
- `BOOT_MODE=psexe`: completely disables both the shell and the CD-ROM booting code, only running the executable specified by `EMBED_PSEXE`. This option is enabled by default for 573 builds but can also be used independently.
- `SPLASH_SCREEN=true`: if enabled, the kernel will draw some color bars as a way of indicating the shell or embedded executable is being loaded, in a similar way to the official 573 BIOS. This should work on non-573 builds as well and is useful if large binaries are embedded into the kernel, as copying them to RAM will take a few seconds.

A System 573 build of OpenBIOS (with a custom shell) has been tested by @NaokiS28 on real hardware and seems to work properly so far. I've also fixed up the Makefile to properly clean all built files regardless of whether or not the options used to run the last build are passed to `make clean`.


src/mips/openbios/main/main.c
```diff
@@ -75,8 +80,9 @@ int main() {
     POST = 0x0f;
     muteSpu();
 
-    if (checkExp1PreHookLicense()) runExp1PreHook();
+    runExp1PreHook();
```
nicolasnoble, line 83: Maybe add a comment here that this is a change from the retail bios in that the check is folded in the function, whereas in retail it's here. Don't forget the code is also meant as a reference to how the retail code is laid out, so we need to comment deviations like that.

src/mips/openbios/main/main.c
```diff
@@ -341,12 +349,14 @@ static void boot(char *systemCnfPath, char *binaryPath) {
...
     startShell(7);
+
+#ifndef OPENBIOS_BOOT_MODE_NO_CDROM
     POST = 8;
     IMASK = 0;
     IREG = 0;
     initCDRom();
     SETJMPFATAL(0x399);
-    if (checkExp1PostHookLicense()) runExp1PostHook();
+    runExp1PostHook();
```
nicolasnoble, line 359: Same here.

nicolasnoble (approved):

PR #1732, opened 2024-09-10
Update vector.hh add consts

mina-skunk:


nicolasnoble:
Sorry for the late reply, was stuck in re-installing my machine and all of its software.

I've amended the PR a bit, which causes API changes to it, but the gist is:

- We shouldn't introduce a `using namespace` at the top of a public library header, as it pollutes the environment of everyone using it. It's best to keep it localized to where it's relevant.
- Using `const` the way you did will create actual variables, which will end up (a) taking memory, and forcing the compiler to do runtime copies of them and (b) if declared in the .hh file like that, can end up creating duplicated declarations. Changing to `static constexpr` solves this.
- The C++-ism of using `static constexpr` within a class means we can't use variables directly. They *need* to be function calls unfortunately. Which means I had to change the API slightly, creating the following usage syntax instead: `psyqo::Vec2::ONE()`.

nicolasnoble (approved):

PR #1767, opened 2024-10-06
Add Icache flush line on clear from memory

marcovoc:
Hello,

I'm working on some hack on spyro 2 pal version, and like it's pointed in #650 the game crash before the main menu with interpreted mode.

It's seem to be the fault of the cache not updated correctly, here my solution to this problem.
If you have a better idea on how to avoid the cache outdated problem, i'm curious to have feeedback.

This code fix the problem spyro 2 pal version success to display the main menu.

PS: it's my first open source contribution, i have no idea of what i'm doing.


nicolasnoble:
Oh I see what you're doing, and it's indeed something I overlooked. Thanks for doing this.

PR #1768, opened 2024-10-07
Fix CDRomDevice nullptr access

eliasdaler:
Asserts were checking CDRom device state on m_device member which was not yet set via registerMe call.

Add isIdle method to check if CDRomDevice is idle instead.

### DIFF
--- a/src/mips/psyqo/cdrom-device.hh
+++ b/src/mips/psyqo/cdrom-device.hh
@@ -321,6 +321,15 @@ class CDRomDevice final : public CDRom {
         S getState() const { return static_cast<S>(m_device->m_state); }
     };
 
+    /**
+     * @brief Checks if the CDROM device is in idle state.
+     *
+     * @details This assumes that all CDRomDeviceStateEnum have a first enum member
+     * with value 0 which corresponds to the "idle" state
+     *
+     */
+    bool isIdle() const;
+
   private:
     void switchAction(ActionBase *action);
     void irq();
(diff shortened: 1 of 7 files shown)


src/mips/psyqo/cdrom-device.hh
```diff
@@ -321,6 +321,15 @@ class CDRomDevice final : public CDRom {
         S getState() const { return static_cast<S>(m_device->m_state); }
     };
 
+    /**
+     * @brief Checks if the CDROM device is in idle state.
+     *
+     * @details This assumes that all CDRomDeviceStateEnum have a first enum member
+     * with value 0 which corresponds to the "idle" state
```
nicolasnoble, line 328: Note that the [concept ](https://github.com/grumpycoders/pcsx-redux/blob/main/src/mips/psyqo/cdrom-device.hh#L47-L55) enforces that already, so the comment is a tiny bit disconnected from the reality.

src/mips/psyqo/cdrom-device.hh
```diff
@@ -321,6 +321,15 @@ class CDRomDevice final : public CDRom {
...
+     *
+     */
+    bool isIdle() const;
```
nicolasnoble, line 331: We might just want to have the implementation here, since it's a simple direct access. Saves a jump and results in the same amount of generated code overall.

nicolasnoble (approved):

PR #1779, opened 2024-10-14
VS Code extension: update to reflect changes in ps1-bare-metal repo

spicyjpeg:
A small update to the VS Code extension that updates and fixes the CMake templates so that they can be built with the latest `ps1-bare-metal` commit, and adds the appropriate code to automatically set up the Python virtual environment required in order to run the tools in the repo. As the virtual environment is added to `.gitignore` by default, a manual "Restore Python environment" option has also been added.


nicolasnoble:
Is the removal of the *.dat files intended? They still seem referenced by the template.

spicyjpeg:
I tweaked the CMake scripts in the last commit to generate `fontImage.dat`and `fontPalette.dat` from `assets/font.png` as part of the build, since having the virtual environment properly configured makes it possible to run `convertImage.py` without needing to manually install any dependencies. I did try building a fresh project generated by the extension and it seemed to work fine, so I shouldn't have missed anything.

That said, I am going to look more closely at how to better detect whether a valid Python installation is present on Windows, as the `py.exe` launcher is apparently not guaranteed to be present or listed in PATH (see the recent discussion on the PSX.Dev Discord server).

Issue #1793, opened 2024-11-04
VSCode Extension does not install MIPS Toolchain on MacOS


rclement:
### Describe the bug

Trying to install the MIPS Toolchain through the PSX.Dev VSCode Extension v0.3.8 on MacOS seems to be working well but keeps failing as not being installed (cannot see the final terminal output as it is auto-closing).

The `mipsel-none-elf-binutils` Brew Formula installs properly though.
All other tools seems to be installed and recognized fined.

Here is the state of the Tools tab after multiple install trials:

<img width="686" alt="Screenshot 2024-11-04 at 23 16 25" src="https://github.com/user-attachments/assets/cd885fcd-a39e-491b-b831-b4d7e34fdc69">

### Expected behavior

The MIPS Toolchain is installed and recognized in the Tools tab (and the template creator enable the "create" buttons).

### Steps to reproduce the bug

1. Install PSX.Dev VSCode Extension
2. Show Panel
3. Go to Tools tab
4. Try to install MIPS Toolchain

### Operating System

MacOS 14.7 (Sonoma)

### PCSX-Redux version

Version: bc029a7f
Build: 107
Changeset: bc029a7f5ef5f0eb6b9baf2ac2a805b948638dd0
Date & time: 2024-10-24 07:41:10

### CPU model

2,7 GHz Dual-Core Intel Core i5

### GPU model & Drivers

Intel Iris Graphics 6100 1536 MB

### BIOS version

OpenBIOS

### Options

- [ ] Dynarec CPU
- [ ] 8MB
- [ ] OpenGL GPU
- [ ] Fastboot
- [ ] Debugger

### Additional information

After investigating the issue, I may have found the culprit in the VSCode Extension implementation:

- The MIPS Toolchain command checks for `mipsel-none-elf-g++` and `mipsel-linux-gnu-g++` executable to be runnable: however both are not found in bin path after the install process
https://github.com/grumpycoders/pcsx-redux/blob/bc029a7f5ef5f0eb6b9baf2ac2a805b948638dd0/tools/vscode-extension/tools.js#L559
- In the `installToolchain` command implementation for MacOS, the `gccScriptPath` variable actually points to the `mipsel-none-elf-binutils.rb` formula instead of the `mipsel-none-elf-gcc.rb` one!
https://github.com/grumpycoders/pcsx-redux/blob/bc029a7f5ef5f0eb6b9baf2ac2a805b948638dd0/tools/vscode-extension/tools.js#L136

[...]

PR #1794, opened 2024-11-04
Fix install of MIPS Toolchain using Homebrew

rclement:
Fixes issue #1793


nicolasnoble:
Oh wow... I'm surprised this is the first time this is actively reported. Thank you for this.

nicolasnoble (approved):

PR #1801, opened 2024-11-16
Patch System fixes/improvements

rocketz:
- Fixed crash when deleting patches from Assembly view
- Added "are you sure?" popup when clicking "delete all"


src/core/patchmanager.h
```diff
@@ -48,7 +48,7 @@ class PatchManager {
     Patch& getPatch(int index) { return m_patches[index]; }
 
     int registerPatch(uint32_t address, Patch::Type type);
-    PatchManager::Patch::Type findPatch(uint32_t address) const;
+    int findPatch(uint32_t address) const;
```
nicolasnoble, line 51: It might be a bit cleaner to use `std::optional<unsigned>` for this, but that'll do for now.

nicolasnoble (approved):

PR #1842, opened 2025-01-13
Fix null pointer error in a psyqo cube template

jahav:
The first parameter of four parameter method `generateRotationMatrix33` expects a pointer to a `Matrix33`. When a 0 is passed, it is interpreted as a null pointer that is later dereferenced to set the matrix and that causes an error.


nicolasnoble:
Ah, we changed that API at some point. Thanks.

nicolasnoble (approved):

PR #1859, opened 2025-02-18
Make pause key unpause as well

mziab:
Having separate keys for pausing and unpausing the emulator felt unintuitive and tripped me up a few times whilst debugging, so I changed the keybinds. Now the pause key can be used for unpausing as well, which feels much better.


nicolasnoble:
Sure, that's fine. The keybindings at some point would really need to be configurable, but let's do this for now.

nicolasnoble (approved):

PR #1862, opened 2025-02-18
Make clicks in assembly view open the memory editor

mziab:
This changes the behavior when clicking highlighted offsets in the assembly view. We open a memory editor if there was none to avoid some possible user confusion as to why the buttons do nothing. The new behavior is also consistent with how the context menu "Go to in memory editor" actions work.

### DIFF
--- a/src/core/system.h
+++ b/src/core/system.h
@@ -89,7 +89,6 @@ struct JumpToMemory {
     uint32_t address;
     unsigned size;
     unsigned editorNum;
-    bool forceShowEditor;
 };
 struct SelectClut {
     unsigned x, y;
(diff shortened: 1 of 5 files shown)


nicolasnoble:
I need to think about this one a bit more.

mziab:
Is there a scenario in which we don't want a memory editor to open (if it wasn't already) when clicking an offset? Maybe the whole forceShowEditor argument isn't actually needed?

I assume this was done on purpose, but it feels weird that the context menu actions open a memory editor while clicking the offsets doesn't.

Alternatively, since I changed the default value for forceShowEditor, I could just omit the third argument in jumpToMemory calls.

nicolasnoble:
Right, I think we should just omit the whole thing really, and just systematically open. We can always re-introduce it later if we realize some UX would benefit from not always opening, but it's fine to clean up now.

mziab:
Simplified the code. It builds and works for me, at any rate. Please look it over.

nicolasnoble (approved):

PR #1863, opened 2025-02-19
Set Terminal to false in desktop entry

malucard:
Simple change, ran into this issue as AppImageLauncher would not pop up to integrate the AppImage into the system if Terminal is set to true

### DIFF
--- a/resources/pcsx-redux.desktop
+++ b/resources/pcsx-redux.desktop
@@ -2,7 +2,7 @@
 Name=PCSX-Redux
 Exec=/usr/bin/pcsx-redux
 Icon=pcsx-redux
-Terminal=true
+Terminal=false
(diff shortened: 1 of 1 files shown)


nicolasnoble:
Wouldn't that prevent it from displaying the console if running from an actual terminal? I have to admit I haven't really looked into the appimage that deeply myself.

nicolasnoble:
You need to start with `-stdio` to see something there.

malucard:
Oh, just saw the flags in the docs. Then yeah, I do see output in the terminal even with Terminal=false. It's also possible to edit the desktop entry after the fact to turn it back on and add `-stdout` if someone wants that extra window.

nicolasnoble (approved):

PR #1867, opened 2025-02-25
Show previous symbol at top of assembly view

malucard:
Unless the top line is already a label or there's no previous symbol. And if the next label is up close, it gets pushed above it.
This is useful for situating yourself in large functions when you have symbol info available.

Before:
Screenshot_20250225_021419

After:
Screenshot_20250225_022434

Getting pushed up by the next label:
Screenshot_20250225_011110

### DIFF
--- a/src/core/r3000a.cc
+++ b/src/core/r3000a.cc
@@ -516,15 +516,15 @@ std::pair<const uint32_t, std::string>* PCSX::R3000Acpu::findContainingSymbol(ui
                 return nullptr;
             }
         }
-		return &*symBefore;
+        return &*symBefore;
     }
-	return nullptr;
+    return nullptr;
 }
 
 std::string* PCSX::R3000Acpu::getSymbolAt(uint32_t addr) {
     auto symBefore = m_symbols.find(addr);
     if (symBefore != m_symbols.end()) {
-		return &symBefore->second;
+        return &symBefore->second;
     }
-	return nullptr;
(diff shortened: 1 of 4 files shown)


nicolasnoble:
Same comment here as with #1868. Let me know if you'd want to try doing the interval tree refactor now, or rather later. However, seeing this PR here now, I think it'd be better to have #1868 use this new method introduced here? Potentially maybe yeeting the symbols containers into its own object for easier refactor later, into https://github.com/grumpycoders/pcsx-redux/blob/7fe5d0adb2a75f8180526b981d5a85291e07215f/src/core/psxemulator.h#L277

malucard:
So, lower_bound actually does a binary search, not a linear search, and symbols can't overlap, so it sounds like an interval tree would work the exact same as the std::map (which in fact seems to use binary searches for everything). I thought a symbol container object like you said might be good, but it'd be very simple, so I just made it a couple of helper functions in the CPU class, which would still allow the other PR to reuse the same function like you suggested. Also, m_elfSymbolsCache was dead code. Is this good?

nicolasnoble:
Oh, I had glanced over the `lower_bound` thing. Yeah it'd be fairly similar then. And yes, the elf stuff is dead code - I used to have a whole DWARF parser in Redux, which I eventually decided to yeet. I guess I forgot this one piece of cleanup.

nicolasnoble (approved):

PR #1868, opened 2025-02-25
Show symbols in callstack view

malucard:
Before:
Screenshot_20250225_042125
After:
Screenshot_20250225_041840

### DIFF
--- a/src/gui/widgets/callstacks.cc
+++ b/src/gui/widgets/callstacks.cc
@@ -22,31 +22,18 @@
 #include "core/callstacks.h"
 #include "core/psxemulator.h"
 #include "core/system.h"
+#include "supportpsx/memory.h"
 #include "fmt/format.h"
 #include "gui/gui.h"
 #include "imgui.h"
 
-void drawSymbol(unsigned pc) {
-    auto& symbols = PCSX::g_emulator->m_cpu->m_symbols;
-    auto symBeforePc = symbols.lower_bound(pc);
-    if (symBeforePc != symbols.begin()) { // verify there is actually a symbol before pc
-        symBeforePc--;
-        if (symBeforePc->first != pc) {
-            uint32_t pcBase =
-                pc >= 0xa0000000? 0xa0000000:
-                pc >= 0x80000000? 0x80000000: 0;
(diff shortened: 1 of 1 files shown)


nicolasnoble:
Couple of comments on this one. I've recently added https://github.com/grumpycoders/pcsx-redux/blob/main/src/supportpsx/memory.h which might be better suited to decode the address for the pcBase stuff. This could simply be:

```c++
PSXAddress pcAddr(pc);
PSXAddress symbolAddr(symBeforePc->first);
if (pcAddr.segment != symbolAddr.segment) { return; }
```
This one's probably simple and straightforward to do here.

The second one might be a bit more involved, but the codebase has something called an [interval tree](https://github.com/grumpycoders/pcsx-redux/blob/main/src/support/tree.h). The associated change would touch a lot of code, but also a lot more beneficial for the general performance of the code there, since it can find the right symbol much faster instead of doing a linear search, which can quickly prove large in some binaries. It's fine if you don't want to do this refactor in this PR of course, given the scope creep.

malucard:
Rebased and updated with the suggestions

malucard:
Alright it's fine now

nicolasnoble (approved):

nicolasnoble:
Oh, this actually fails to build on Windows. I'll check what's going on...

src/gui/widgets/callstacks.cc
```diff
@@ -22,10 +22,21 @@
...
 #include "gui/gui.h"
 #include "imgui.h"
 
+static void drawSymbol(unsigned pc) {
+    std::pair<const uint32_t, std::string>* symbol = PCSX::g_emulator->m_cpu->findContainingSymbol(pc);
+    if (symbol) {
+        ImGui::SameLine();
+        ImGui::TextUnformatted(" :: ");
+        ImGui::SameLine();
+        ImGui::TextUnformatted(symbol->second.data(), &*symbol->second.end());
```
nicolasnoble, line 36: That's actually not valid. You can't dereference `end()`, it causes an assert. I'm making changes on my end.

nicolasnoble (approved):

PR #1871, opened 2025-02-26
Fix build with fmt 11

malucard:
Needed this to build on Arch Linux


nicolasnoble (approved):
Hah, the amount of time fmt upgrades breaks the build... Thanks.

PR #1872, opened 2025-02-27
Save window maximized state

malucard:
Saves and restores whether the window is maximized together with the other window state data.

(Note: it's important that the size not be set if the window is to be maximized, that caused it to get stuck in the maximized size without decorations for me)


nicolasnoble (approved):
Thank you. This was on my todo-list, but low enough that I never got to it.

PR #1873, opened 2025-02-28
Add intuitive dockspace

malucard:
This adds a dockspace to the entire window that switches the full window render option on and off automatically. Specifically:
When full window render is on, docking anything onto the dockspace will cause full window render to be disabled, and the output window will be placed in the remaining dock space.
When full window render is off, if the output window is the only visible window docked on the dockspace and there are no other tabs, full window render will be enabled and any hidden windows on the dockspace will be undocked.
The toggle still works as before, it undocks everything when you click it in order to bypass this system.

What is now possible:
image


malucard:
Also made the output window display centered because that seems like the right thing to do

nicolasnoble (approved):

PR #1886, opened 2025-03-11
[Chores] Format code

nicolasnoble:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


rixnobis:
Closing by deleting the branch, so the job can regenerate it.

`code-format` and `regen-i18n` have both failed since 2026-09-18, at the `Create Pull Request` step, with `error: The following untracked working tree files would be overwritten by checkout`. The cause is this branch: it carries tracked copies of paths that are submodules in the current tree, so checking it out collides with the submodule checkouts. On 2026-09-19 the collision was `third_party/iec-60908b/*`; since #2112 it is `src/mips/*` as well.

Nothing else fails in either run, so deleting the branch clears it and the next push to main will open a fresh one against the current layout. The content here is a pass over a tree that no longer exists, so there is nothing to carry across.

PR #1891, opened 2025-03-13
Stabilizing the order of symbols and sections + various other fixes.

### LINKED ISSUE #1890: psyq-obj-parser giving different output (release vs local)
It appears compiling `psyq-obj-parser` locally and using it generates a different output object file than using the version included in the latest release zip. I'm compiling with this command:
```
g++ -o psyq-obj-parser.exe -std=c++2b -Ithird_party -Ithird_party/ELFIO -Ithird_party/fmt/include -Ithird_party/magic_enum/include/magic_enum -Isrc third_party/fmt/src/os.cc third_party/fmt/src/format.cc -static -O2 tools/psyq-obj-parser/psyq-obj-parser.cc src/support/file.cc
```
Compiling with mingw32, mingw64, and ucrt64 all generate the same output, which is the correct output. This creates and elf file where all sections and data are in the same order as the original input object. Using the version inside the release zip however, results in a fairly random/scrambled output.

Example, using the locally built version (correct):
```
  :: Generating sections
    :: Generating section .rdata
    :: Generating section .text
    :: Generating section .data
    :: Generating section .sdata
    :: Generating section .sbss
    :: Generating section .bss
```
Using the release version (incorrect):
```
  :: Generating sections
    :: Generating section .sbss
    :: Generating section .sdata
    :: Generating section .bss
    :: Generating section .rdata
    :: Generating section .data
    :: Generating section .text
```
I can only think that this is either due to me missing some kind of compiler flag, which ends up actually fixing the bug, or that there is undefined behavior somewhere which results in different handling for Visual Studio vs GCC.

nicolasnoble:
Fixes #1890.

### DIFF
--- a/src/support/list.h
+++ b/src/support/list.h
@@ -147,7 +147,9 @@ class List final {
     bool empty() const { return m_count == 0; }
     void clear() {
         for (Node* ptr = m_head.m_next; ptr; ptr = ptr->m_next) ptr->m_parent = nullptr;
-        m_head.m_next = m_tail.m_prev = nullptr;
+        m_head.m_prev = m_tail.m_next = nullptr;
+        m_head.m_next = &m_tail;
+        m_tail.m_prev = &m_head;
         m_count = 0;
     }
     iterator insert(iterator i, Node* node) {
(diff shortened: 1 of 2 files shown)


galaxyhaxz:
Just want to confirm it is now producing the same results with both builds, and when using the new `-s` flag, is creating the proper binary exact elf files. Thanks a bunch for taking the time to fix this :)

Issue #1681, opened 2024-07-04
When increasing the interface font, memory editor jumps to wrong location


marco-calautti:
### Describe the bug

I tried increasing the interface font size to 16. However this makes the Memory editor visually jump to the wrong location when entering an address and pressing enter.

### Expected behavior

The UI location where the cursor jumps in the memory editor should adapt to the font size.

### Steps to reproduce the bug

1. Change the font size to 16 (both fonts)
2. Enter an address I. The memory editor and press enter.

### Operating System

Arch Linux

### PCSX-Redux version

Latest AppImage

### CPU model

i7

### GPU model & Drivers

gtx 1650 ti

### BIOS version

SCPH1001

### Options

- [ ] Dynarec CPU
- [ ] 8MB
- [ ] OpenGL GPU
- [X] Fastboot
- [X] Debugger

PR #1905, opened 2025-03-15
Fix invalid set scroll position in memory editor.

ffgriever-pl:
If a monospace font is used for memory contents, use its size to calculate jump-to position. Previously, if monospace and variable width font sizes were different, the set scroll position would be invalid (point to offset before or after selected one).


ffgriever-pl:
Fixes #1681

nicolasnoble (approved):

PR #1907, opened 2025-03-17
psyq-obj-parser: Handle comm symbols properly

dezgeg:
Convert .comm symbols into the proper ELF equivalent, instead of converting them to local symbol in the .bss segment.

This makes it possible use the converted PSY-Q libraries to recreate a bit-exact executable of Frogger - without proper comm symbol support controlling the order of some .bss symbols in the libraries is impossible.

Even though this is more correct than before there are some known users of psyq-obj-parser are relying on the old behaviour thus a command line option is added which can be used to restore the previous behaviour.

(Alternate variant at https://github.com/grumpycoders/pcsx-redux/compare/main...dezgeg:psyq-obj-parser-comm-support-variant1 but I think this one looks clearer).


nicolasnoble:
Bit busy this week, I'll look at the various PRs over the week-end.

galaxyhaxz:
This is technically the proper way to handle these and what GCC would normally generate. However, for in-progress game decompilations, tools like splat expect `.sbss/.bss`, whereas you must substitute this for `.scommon/COMMON` instead. I only ran into one problem with this change, it no longer keeps the bss variables in the correct order, in fact, uses the same order regardless if the `-s` option is used. Might be good to have this as a command line option, just so the existing functionality can stay to force variables into the correct order and sections.

Kneesnap:
> Might be good to have this as a command line option, just so the existing functionality can stay to force variables into the correct order and sections
There were 2 other branches linked in the first post which made it available behind a command-line flag. One of them should probably be merged instead of the branch assigned to the PR in order to maintain compatibility with existing projects which don't migrate to this more accurate design.

>That is, with this change, regardless of what order a variable appears in COMM, it seems to link in a completely different order, whereas when they are stored in BSS, they link in the same order they appear in the ELF)

This is intended. A translation unit using common symbols do not have the ability to specify the ordering of common symbols by design. Instead, it's the linker's responsibility to order them. This behavior is consistent with the original ASPSX.EXE + PSYLINK.EXE behavior.

That may sound like you're losing functionality, but it's actually the opposite: this gives more control now. But the responsibility for using this properly is now on each decomp project instead of psyq-obj-parser. The advantage to this is that symbols from the same object don't need to be grouped together anymore, they can be freely moved around, even amongst the common symbols of other objects.

This can be done by creating a single .S file which defines all common symbols from all translation units, in whichever order you'd like. ([Example](https://github.com/HighwayFrogs/frogger-psx/blob/linux-support/source/linux/bss.s))

galaxyhaxz:
> There were 2 other branches linked in the first post which made it available behind a command-line flag.

Yeah I saw those, that's why I made the suggestion it would be better to go with one of those. Once a given project is finally in a "shiftable" state, then symbol order isn't really a problem. But in the long interim of reversing a game, I find it super helpful to be able to finish a file, have all functions/variables in it and moved out of the asm stuff, and then compile that file to an OBJ, convert to elf, and produce a bin-exact executable. Of course, once this has been done for every file and they can finally be linked with psylink rather than converting to elfs, they will likely scramble again as you mention, the ordering is done by hash, so it would be extremely hard to get them right short of just guessing/appending junk onto names to get the hash in the right order.

dezgeg:
Thanks for testing!

> I only ran into one problem with this change, it no longer keeps the bss variables in the correct order, in fact, uses the same order regardless if the `-s` option is used.

Can you upload an object file where this happens somewhere, maybe I accidentally broke something unrelated.

galaxyhaxz:
You technically didn't break anything, as @Kneesnap already mentioned, this is the correct/expected behavior of communal definitions. Multiple files could technically reference the same definition, but the linker has the freedom to move them around and place them how it wants. This is how an obj is properly converted, but has the downside that when decompiling, you often dont have the original variable names, and since COMMs appear to be sorted by hash, this makes it impossible to get them in the right order. Whereas, when they sit within the bss sections, it simply "copies" the bss section out of the elf as a solid block, so everything stays in the same order even if the names are incorrect.

nicolasnoble:
Sorry, got hit hard by covid. I'm processing my backlog slowly now.

dezgeg:
No problem!

Kneesnap:
bump again

PR #1908, opened 2025-03-18
Add Take Screenshot option

mziab:
This adds a Take Screenshot option to the Emulation menu.

I want to stress that this is a tentative implementation and I would like to request comments on the following:
1. I have copied the convertScreenshotToImage and writeImagePNG helpers from the webserver to PCSX::GUI (without removing the originals). Since now there are two consumers for these, what would be the proper place to put them?
2. Should the helpers and the new getDateString one be renamed to something else? Additionally, is PCSX::GUI the proper place for the date string helper?
3. Maybe a hotkey should be added for this? I was thinking of F12, like Steam uses.


nicolasnoble:
Bit busy this week, I'll look at the various PRs over the week-end.

nicolasnoble:
Sorry, took a while to get back to you, lots of real life stuff.

So, yeah, it might be relevant to create a new file, like screenshot.h/screenshot.cc to factorize this a bit. We can create a few static functions in a namespace like `PCSX::ScreenShot`.

For the date helper, it probably would belong into the screenshot namespace for now? We can move it if needs be if we start re-using it elsewhere maybe?

For the hotkey, sure, I don't mind. Note there's already a hotkey to take a screenshot into the clipboard, and the code probably needs to be refactored with the new screenshot code too: https://github.com/grumpycoders/pcsx-redux/blob/34ea4f188bd4e0630e6f43bfbcd0060f96704310/src/gui/gui.cc#L1095-L1125

PR #1911, opened 2025-03-22
Adding command buffer and test command

johnbaumann:


src/mips/psyqo/hardware/cdrom.hh
```diff
@@ -2,7 +2,7 @@
 
 MIT License
 
-Copyright (c) 2023 PCSX-Redux authors
+Copyright (c) 2025 PCSX-Redux authors
```
nicolasnoble, line 5: Don't bump copyrights unless there's a fundamental rewrite of the file.

johnbaumann: Ah, I thought it was just on updates

nicolasnoble: I've been following the guidance of an OSS lawyer from a decade ago, basically, who said that copyright date of a single file starts when the file is first created, and the date should only bump when the file is fundamentally being rewritten. The reasoning is this is to know when the copyright actually expires, ala Mickey Mouse.

src/mips/psyqo/cdrom-commandbuffer.hh
```diff
@@ -0,0 +1,55 @@
...
+
+*/
+
+#pragma once
+
+#include <stdint.h>
+
+#include <concepts>
+#include <type_traits>
+
```
nicolasnoble, line 33: We're missing a namespace here. Probably `psyqo::Hardware::CDRom`

src/mips/psyqo/cdrom-device.hh
```diff
@@ -412,6 +413,15 @@ class CDRomDevice final : public CDRom {
      */
     void setVolume(uint8_t leftToLeft, uint8_t rightToLeft, uint8_t leftToRight, uint8_t rightToRight);
 
+    /**
+     * @brief Sends a test command to the CDRom mech
+     *
+     * @param callback The callback to call when the command operation is complete.
+     */
+    void test(CDRomCommandBuffer commandBuffer, eastl::function<void(bool)> &&callback);
```
nicolasnoble, line 421: This probably should be a const reference, for all 3.

src/mips/psyqo/src/cdrom-device-test.cpp
```diff
@@ -0,0 +1,85 @@
...
+enum class TestActionState : uint8_t {
+    IDLE,
+    TEST,
+};
+
+class TestAction : public psyqo::CDRomDevice::Action<TestActionState> {
+  public:
+    TestAction() : Action("TestAction") {}
+
+    void start(psyqo::CDRomDevice *device, CDRomCommandBuffer commandBuffer, eastl::function<void(bool)> &&callback) {
```
nicolasnoble, line 46: This CDRomCommandBuffer should be a const reference.

src/mips/psyqo/src/cdrom-device-test.cpp
```diff
@@ -0,0 +1,85 @@
...
+}  // namespace
+
+void psyqo::CDRomDevice::test(CDRomCommandBuffer commandBuffer, eastl::function<void(bool)> &&callback) {
+    Kernel::assert(m_callback == nullptr, "CDRomDevice::test called with pending action");
+    s_testAction.start(this, commandBuffer, eastl::move(callback));
+}
+
+psyqo::TaskQueue::Task psyqo::CDRomDevice::scheduleTest(CDRomCommandBuffer commandBuffer) {
+    return TaskQueue::Task(
+        [this, commandBuffer](auto task) { test(commandBuffer, [task](bool success) { task->complete(success); }); });
```
nicolasnoble, line 75: So, this one will allocate memory implicitly, which isn't something we want. What we should be doing instead here is the following:

- Create a CDRomCommandBuffer inside the s_testAction.
- Copy the buffer to it.
- Add a new `start` override without the buffer as a parameter, which should just call into the normal `start` version.
- Call `start` from within the task's lambda instead.

This way there's no need to capture the command buffer inside the task's lambda, and the action will hold it statically.

nicolasnoble (approved):

src/mips/psyqo/examples/cdrom-demo/cdrom-demo.cpp
```diff
@@ -75,6 +76,11 @@ void CDRomDemo::createScene() {
             syscall_puts("Failure\n");
         }
     });
+
+
+    psyqo::Hardware::CDRom::CDRomCommandBuffer commandBuffer;
+    commandBuffer.set(0x50, 0xf2, 0xde, 0xad);
+    m_cdrom.testBlocking(gpu(), commandBuffer);
```
johnbaumann, line 83: I didn't realize I committed this change to the repo - is there a clean way to roll back this change or prevent it from getting merged in? Otherwise I can just revert the changes.

nicolasnoble: Eh, no need to rewrite history for this. Just push a commit to remove it if you want it gone.

nicolasnoble (approved):

PR #1925, opened 2025-04-13
Add turbo hotkey

mziab:
This adds a simplistic turbo hotkey which toggles the speed scaler to 200% speed and back. This is a tentative implementation with one problem I haven't been able to fix.

The main issue is that setting the speed scaler back to 100% occasionally makes the emulator hang (which also happens when moving the speed scaler slider left, even without this patch). Probably the tick counter getting out of alignment? Is there a better way of implementing turbo mode I should consider? I'd really appreciate some guidance on this.

Aside from that, both the speed and hotkey are hardcoded for now.


nicolasnoble:
So I need to rewrite the synchronization system between the CPU and the SPU, which is how the speed system currently works, and the reason messing with it too much in real time causes these hiccups. It's not exactly difficult to rewrite, really.

Right now, the scale system changes this value, which can cause things to go desync when changing it while it's running:

https://github.com/grumpycoders/pcsx-redux/blob/main/src/core/psxcounters.cc#L147

The right way to do it is instead to compress the audio stream here according to the scaling:

<https://github.com/grumpycoders/pcsx-redux/blob/main/src/spu/miniaudio.cc#L181-L220>

As for the details of the implementations, it might be better to currently avoid changing the scaler value, and instead add a getter for it, multiplying it by the turbo factor when the key is pressed. This way, (1) the setting itself doesn't change, and (2) it'll be easier to move the code over to the right spot.

PR #1948, opened 2025-08-01
VS Code extension: fix CMake installation and Python venv setup on Windows

spicyjpeg:
Fix for the `createPythonEnv()` function in the PSX.Dev VS Code extension, which is currently broken on Windows due to it not taking into account the "fake" Python executables that redirect to the Microsoft Store as well as virtual environments having a slightly different file structure depending on the platform. I have also fixed `installCMake()` and bumped the extension's version number to 0.3.10.

**I haven't yet tested these changes on a Windows machine so please review before merging.**


nicolasnoble:
Folded into #1966, with your CMake and Python fixes kept and credited as co-author. Thanks!

PR #1953, opened 2025-08-06
nix: Fix submodules

cleverca22:
this PR switches the flakes from using submodules manually (which can also break due to https://github.com/NixOS/nix/issues/13698 )
over to using submodules via flakes (which then also needs nix to be upgraded)

this is required, because fetching the project via a flake URI such as `nix build github:grumpycoders/pcsx-redux/1b0cbe5e4a535def2e3e3279551e6665cdc102ae` completely breaks the build
which also makes it impossible to add this project as an input on another flake


nicolasnoble:
@NotExactlySiev ping

NotExactlySiev:
Your solution works perfectly (thanks!) and could be merged in right now. However: those `cp` commands were meant to be only a temporary solution until I figure out a better way to do it. The correct solution imo would be to stop having those dependencies' paths be hardcoded to the `third_party` directory and instead modify the Makefile to receive their paths from an environment variable set by the derivation. This way the build can access these dependencies directly from the store without needing to copy anything. We already did something a bit like this in the Makefile for the luajit deps.

Also: I added `NIX_BUILD_CORES = 4` because trying to build redux on my (16GB RAM + 16GB Swap) laptop causes the entire PC to hang for a few minutes before the build process crashes :P But I'm fine with removing it as I don't want other people who unlike me don't have limited RAM be forced into slow builds. Ideally there could be a check inside the derivation for the amount of available RAM before setting that variable, but I don't know how something like that will ever work within Nix's purity limits. 
For now we can have it be unlimited by default and I'll just override it on my local system. But I expect a few other users to eventually run into the same problem and be confused as to why attempting to install pcsx-redux hangs their nix system :stuck_out_tongue: So that's something we might have to think about.

cleverca22:
> Also: I added NIX_BUILD_CORES = 4 because trying to build redux on my (16GB RAM + 16GB Swap) laptop

that should really be set in `nix.conf`
the way it is currently, it actively ignores my config, and i have to change the source to build with a different count
and because of the nix fetching bug, any change of the source fixed the problem, making it very hard to tell if it was a concurrency bug, due to it always using 4 threads

NotExactlySiev:
We're still mostly using the vendored dependencies. it's just that nix likes to download them to `/nix/store/<some random string>` which then has to be copied into `third_party`. it would be nice if the build was able to address them where they're downloaded so we can avoid the copying.

PR #1956, opened 2025-08-13
a64 JIT: Fix `setRWX` returning nothing on Apple

wheremyfoodat:
This shouldn't change anything because the function was not being called at all on MacOS, but it's best to avoid any future happy accidents in case that changes.

On a relevant note, I need to fix Windows support for the a64 JIT, but I haven't gotten my WoA dev setup running yet.


wheremyfoodat:
Update on Windows-on-ARM support: Redux doesn't currently build with either the MSVC ARM64 target (Because some dependencies don't have ARM64 DLLs on Nuget) nor with the ARM64EC target (Because both the core and third party code get confused about whether they're compiling for x64 or arm64...), so fixing the JIT there is unlikely to happen any time soon.

Thankfully ARM Linux/MacOS work fine, and x64 builds also work fine on WoA.

nicolasnoble (approved):

PR #1957, opened 2025-08-15
Initial splash screen implementation

wheremyfoodat:
- Displays a static splash screen with usage instructions when the emulator starts
- Adds a Python script for generating splash screen data from an image file
- Fixes glTexParameteri usage. The UI offscreen textures are no longer written to before min/mag filters are set, which would make Apple's drivers error.

The current splash screen is a sample and not necessarily final. This code will also likely be replaced when nanovg/nanosvg are replaced.

https://github.com/user-attachments/assets/f82813ee-b31e-47bd-8d2a-fb16f3428588

Image used for the splash screen in the video (Made in Canva, so I sadly can't share the project files easily)
redux-splash


nicolasnoble (dismissed):

nicolasnoble:
Looks like this breaks nix.

nicolasnoble (approved):

PR #1973, opened 2025-10-22
Remove deprecated fmt::localtime API and update fmt to 12.0.0

### LINKED ISSUE #1969: Compilation with fmt 12.x fails
### Describe the bug

`fmt::localtime` was removed after deprecation in previous versions, in favor of `std::localtime`.

I noticed because Arch Linux upgraded to fmt 12.x yesterday (but afaiu, fmt is pulled in as a git submodule for this project)

### Expected behavior

For `pcsx-redux` to compile with fmt 12.x

### Steps to reproduce the bug

1. Try to compile
2. Observe fmt::localtime compilation errors

### Operating System

Arch Linux

### PCSX-Redux version

main/master

### CPU model

Intel Core 13900T

### GPU model & Drivers

NVIDIA 4090

### BIOS version

SCPH1001

### Options

- [ ] Dynarec CPU
- [ ] 8MB
- [ ] OpenGL GPU
- [ ] Fastboot
- [ ] Debugger

### Iso checks

_No response_

### Logs

_No response_

### Additional information

_No response_

njfox:
Fixes #1969 

`fmt::localtime` was deprecated and finally removed in fmt 12, which was pushed to Arch Linux recently, and this caused build failures and runtime errors when linked against the system fmt library. This PR removes the deprecated function and replaces it with a new `formatTimestamp()` function in `VersionInfo` that formats the timestamp based on a format string argument passed by the caller. 

Note that these changes are _**incompatible**_ with fmt10, so we needed to update the submodule as well.

### DIFF
--- a/src/gui/gui.cc
+++ b/src/gui/gui.cc
@@ -63,7 +63,6 @@ extern "C" {
 #include "core/sstate.h"
 #include "core/web-server.h"
 #include "flags.h"
-#include "fmt/chrono.h"
 #include "gui/gui.h"
 #include "gui/luaimguiextra.h"
 #include "gui/luanvg.h"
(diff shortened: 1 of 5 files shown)


rubin55:
fwiw, I needed this patch to get pcsx-redux to compile again on Arch Linux (using the `pcsx-redux-git` AUR package, which does not use third-party/fmt btw)

nicolasnoble:
Superseded.

PR #1978, opened 2025-11-13
fmt has broken the build again

malucard:
fmt::localtime was removed in favor of std::localtime, so I had to change it to that to be able to build the emulator


nicolasnoble:
Right, bit of an overlap with #1973, but this one is a tad more simple. I'll take it but then I'll probably change it more to what #1973 does after upgrading it.

PR #2018, opened 2026-05-23
Add SPU mips tests (test on HW help needed)

Xeeynamo:
I am experimenting a SPU regression suite on a SPCH-5552 to raise the base line to which emulators should adhere to reproduce PS1 sounds accurately.

As of today, tests are non-exhaustive and they occasionally suffer from synchronization issues. These tests aim to purely focus on ADPCM decoding, resampling and the rate ADSR envelopes change the voice volume given a certain set-up. These tests should not depend to accurate CPU cycles, as it would defeat their purpose.

The PR is in draft. I am mostly requesting for feedback. The SPU synchronization is imperfect, and both ADPCM and voice tests might occasionally fail on real hardware. These tests have not been tested on older or newer PS1 revisions, on a PS2 or arcade boards. Tests hang on PCSX Redux due to the SPU Status bit11 flip not being implemented.


nicolasnoble:
I think this generally speaking looks good. We may want to avoid the `run` target in the Makefile and leave it to the operator to decide how to actually run those, given there's multiple harnesses around, but that's my only very small nit.

Xeeynamo:
> I think this generally speaking looks good. We may want to avoid the `run` target in the Makefile and leave it to the operator to decide how to actually run those, given there's multiple harnesses around, but that's my only very small nit.

Thanks for the feedback. I removed the `run` but kept `spu_dump` for simplicity. Let me know if you want that to be removed as well.

For the tests themselves, I reached a point where I am happy with what I wrote so far. I am happy to make further changes where I missed the mark.

These tests have been tested only on one console. Before I have the confidence to mark the PR as ready, I'd like help to collect the test run results from other consoles as well.

nicolasnoble:
Pushed a few commits to your branch: merged main (the SPUVoice `ad`/`sr` fields got renamed to `adsrLo`/`adsrHi`, so it wasn't building), fixed the SPU_DUMP path (a duplicate `static` plus it was writing uninitialized `warmup`/`period`, so the dumps came out header-only), made the golden compare tolerate the capture-start offset, and disabled interrupts during the measurements.

On the offset: I ran this on a 5501 against your 5552 goldens. The decoded samples are bit-identical, just shifted by one capture tick - the bit-11 sync only pins the start to ~1 sample, and which boundary it lands on differs per boot and per console revision. So a byte-exact compare was never going to be portable; the new one finds the integer-sample alignment first.

Two things still flaky on real hardware, more in your court since they're test-design calls:

- The ADSR rate tests use a ±2 envx tolerance, which can't survive ~1 sample of sync jitter at the faster rates - they miss on most boots here. Probably wants a wider band, or to gate the envx reads on the bit-11 edge more tightly.
- The drain/bit-11 sync occasionally misses by more than a sample (~1 in 5-6 boots), throwing the whole capture off. Disabling interrupts helped (got a clean 23/23 once) but didn't eliminate it.

6-boot sweep on the 5501 after these changes: 23/23, 19/23, 19/23, 14/23, 11/23, 21/23 - the sub-19 ones are the gross sync misses, the 19-21 ones are just the ADSR tolerances.

src/mips/tests/spu/spu.c
```diff
@@ -340,6 +340,25 @@ static void run_voice1_with_sample(const uint8_t *sample64, uint16_t pitch) {
...
+static int s_interruptsWereEnabled;
+)
+
+// The bit-11 capture sync and the busy-wait-timed envelope reads are both
+// jitter-sensitive: an interrupt (Unirom's SIO/timer/vblank handlers) firing
+// mid-measurement perturbs the timing, mis-syncing a capture or pushing an
+// envx read past its tolerance. Disable interrupts for the suite, as the
+// cop0/cpu/timer tests do.
+CESTER_BEFORE_ALL(spu_tests,
+    s_interruptsWereEnabled = enterCriticalSection();
```
Xeeynamo, line 353: @nicolasnoble very clever, thanks!!

rixnobis:
Closing this, and saying where it went.

These tests are in the `spu-accuracy` branch, #2077. Thirteen of the seventeen files here are carried there byte-identical - all eleven `.pcm` hardware goldens, `spu-adpcm.c` and `spu-transfer.c` - and the Makefile, `spu.c`, `spu-adsr.c` and `spu-capture.c` were extended from yours. That happened without your name attached. The omission was mine. #2077 now credits you in its body and carries a `Co-authored-by`.

Separately, #2112 moves `src/mips` out to `pcsx-redux/nugget` and has redux consume it as a submodule, so the path this PR targets is about to stop existing in this repository. Future SDK work goes to nugget.

On the "test on HW help needed" in your title: I have the hardware, including several consoles and a devkit. If there is anything in the SPU you wanted measured on real silicon and could not reach, open an issue on `pcsx-redux/nugget` and I will run it and post the captures.

PR #2023, opened 2026-05-25
Improved register view

rocketz:
* Now shows both hexidecimal and decimal at the same time
* Color coded registers by type for easier overview
* Inline editing of values - both as hex or decimal
* Keyboard navigation

### DIFF
--- a/src/gui/widgets/registers.cc
+++ b/src/gui/widgets/registers.cc
@@ -55,19 +55,22 @@ void PCSX::Widgets::Registers::contextMenu(const char* name, uint32_t reg, uint3
 
 void PCSX::Widgets::Registers::updateRegisterValue(PCSX::psxRegisters* registers, const std::string& val, int reg,
                                                    int radix) {
+    if (val.empty()) return;
+    char* endPtr = nullptr;
     uint32_t newReg;
-    char* endPtr;
     if (radix == 16) {
-        newReg = strtoul(val.c_str(), &endPtr, 16);
+        uint64_t parsed = strtoull(val.c_str(), &endPtr, 16);
+        if (*endPtr || parsed > UINT32_MAX) return;
+        newReg = (uint32_t)parsed;
     } else {
-        newReg = strtoll(val.c_str(), &endPtr, 10);
+        int64_t parsed = strtoll(val.c_str(), &endPtr, 10);
+        if (*endPtr || parsed < INT32_MIN || parsed > UINT32_MAX) return;
+        newReg = (uint32_t)parsed;
     }
-    if (!*endPtr) {
-        if (reg == 34) {
-            registers->pc = newReg;
-        } else {
-            registers->GPR.r[reg] = newReg;
-        }
+    if (reg == 34) {
+        registers->pc = newReg;
+    } else {
+        registers->GPR.r[reg] = newReg;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/gui/widgets/registers.cc
```diff
@@ -40,17 +41,33 @@ void PCSX::Widgets::Registers::makeEditableRegister(const char* name, uint32_t r
...
+    } else {
+        newReg = strtoll(val.c_str(), &endPtr, 10);
+    }
+    if (!*endPtr) {
+        if (reg == 34) {
+            registers->pc = newReg;
+        } else {
+            registers->GPR.r[reg] = newReg;
+        }
     }
```
nicolasnoble, line 73: This accepts empty input and writes 0, and out-of-range values get silently truncated to uint32. Want an early-return on empty, parse into a 64-bit int, then range-check against UINT32_MAX/INT32_MIN before committing.

```suggestion
    if (val.empty()) return;
    char* endPtr = nullptr;
    uint32_t newReg;
    if (radix == 16) {
        uint64_t parsed = strtoull(val.c_str(), &endPtr, 16);
        if (*endPtr || parsed > UINT32_MAX) return;
        newReg = (uint32_t)parsed;
    } else {
        int64_t parsed = strtoll(val.c_str(), &endPtr, 10);
        if (*endPtr || parsed < INT32_MIN || parsed > UINT32_MAX) return;
        newReg = (uint32_t)parsed;
    }
    if (reg == 34) {
        registers->pc = newReg;
    } else {
        registers->GPR.r[reg] = newReg;
    }
```

src/gui/widgets/registers.cc
```diff
@@ -83,13 +112,53 @@ void PCSX::Widgets::Registers::draw(PCSX::GUI* gui, PCSX::psxRegisters* register
...
+                ImGui::TextColored(colors[colorIndices[counter]], "%s:", name);
+                ImGui::SameLine();
+
+                char label[8];
+                std::snprintf(label, sizeof(label), "##%2d", counter);
+                std::string val = fmt::format("{0:08x}", reg);
+                if (ImGui::InputText(label, &val,
+                                     ((counter == 0) ? ImGuiInputTextFlags_ReadOnly : 0) |
+                                         ImGuiInputTextFlags_CharsHexadecimal | ImGuiInputTextFlags_EnterReturnsTrue |
+                                         ImGuiInputTextFlags_AutoSelectAll)) {
```
nicolasnoble: EnterReturnsTrue drops the edit on tab or click-away, only Enter commits. IsItemDeactivatedAfterEdit() commits on focus-loss instead. Worth landing after the parser fix above, otherwise auto-commit makes accidental zeroing easier to hit. Same one-liner at the other 3 InputText sites: GPR decimal, GPR pc, Misc pc.

```suggestion
                                         ImGuiInputTextFlags_AutoSelectAll) ||
                    ImGui::IsItemDeactivatedAfterEdit()) {
```

src/gui/widgets/registers.cc
```diff
@@ -23,14 +23,15 @@
 #include <numbers>
 
 #include "core/disr3000a.h"
+#include "core/psxmem.h"
```
nicolasnoble: core/psxmem.h looks unused now, drop it.

nicolasnoble (commented):
Moving these to inline editable fields is the right call over the modal popups. Two coupled things on the input path, see the inline notes.

nicolasnoble (approved):
Input-path notes are all addressed, this is good to go. Thanks!

PR #2042, opened 2026-06-23
Adding memory card support to psyqo

nicolasnoble:


src/mips/common/util/sjis-encode.hh
```diff
@@ -0,0 +1,115 @@
+#pragma once
+
+#include <stdint.h>
+
+#include "common/util/sjis-encode-table.h"
+
+// Unicode -> Shift-JIS encoding, the inverse of support/sjis_conv.cc's decoder.
```
eliasdaler, line 34: Maybe it would be good to add an EASTL version of sjis_conv.h/.cc to psyqo to make it possible to display/printf obtained filenames? 

Right now you can't use it as-is, because sjis_conv.h/.cc use <string> which is not available in freestanding mode.

PR #2060, opened 2026-07-18
Nix: Don't include the mipsel toolchain in the default dev shell.

NotExactlySiev:
Currently someone wanting to work on redux while using nix would have to rebuild the entire mipsel toolchain just to access the default dev environment. Which they probably won't use anyway if they just want to work on the emulator itself. I added a "no openbios" version of redux and change the dev environment to use that one by default.

someone is me. I don't like having to build this specific version of mipsel-gcc every time I wanna compile redux.


NotExactlySiev:
Just realized this has doubled the number of github actions tasks :stuck_out_tongue:

NotExactlySiev:
I'm grouping the rest of the nix related fixed with this. Updating the flake and therefore the mesa dependency, so we don't get the glfw error, etc. No point having two separate PRs for these.

nicolasnoble (approved):

PR #2066, opened 2026-07-29
Factor out the network layer, and give it a window

nicolasnoble:
The GDB server, the web server and the SIO1 pair each hand-rolled the same
bind/listen sequence, each declared its own three-state status enum, and each
dropped every error silently. Nothing read those enums, and both servers carry a
`m_gotError` member that is never written to. So a server that fails to bind
leaves its checkbox ticked and says nothing, and unticking it afterwards trips
`uv_close`'s `!uv__is_closing` assertion.

Everything goes through `UvFifoListener` now, and connections are `IO<File>`.
`Network::Endpoint` carries one status vocabulary and a last error; a concrete
endpoint says what to do with a connection and nothing else, which deletes 191
lines from the GDB server and 147 from the web server.

Three transport fixes were needed to get there:

- `UvFifoListener` reports bind failures instead of closing and returning. This
  is what fixes the abort.
- `UvFifo::close()` flushes pending writes. `uv_close` cancels queued writes, so
  closing right after writing truncated the tail - 64MB written then closed
  delivered 2634240 bytes. The web client's `m_closeScheduled` bookkeeping
  existed to work around this and is now gone.
- A failed connect clears `m_connecting`, which was only cleared on success. The
  SIO1 reconnect button is gated on `!connecting() && fifoError()`, so it could
  never become clickable after the failure it exists to recover from.

`UvFifo` also gains an optional readable notification, so the ported servers
stay event-driven instead of polling at frame granularity like SIO1 does.

The services move out of the emulation configuration window into a Network
window with a status indicator, a restart button and the error string. The SIO1
client host field loses `ImGuiInputTextFlags_CharsDecimal`, which made it
impossible to type a hostname into it.

Unrelated but needed along the way: test objects were built in place by make's
built-in suffix rule, so they had no header dependency tracking and `clean` left
them behind. Editing a header a test includes relinked the stale object, which
is heap corruption at runtime rather than a build error when the header changes
a class layout. They build under `objs/` now.

Tested with new gtests in `tests/support/`, and against running builds: GDB
answers `?`/`qSupported`/`g` correctly, the web server returns full 1MB and 2MB
bodies, SIO1 accepts connections.


nicolasnoble:
The window, with the GDB server and the web server deliberately pointed at the
same port so one of them has to lose:

Network window

PR #2076, opened 2026-08-20
psyqo: GTE-accelerated math, and a teapot to demonstrate it

nicolasnoble:
The GTE's lighting half had no worked example. Everything in examples/ that
shades anything does it on the CPU, so the light source matrix, the light
colour matrix and the nc* commands were entirely undemonstrated - which is
awkward when somebody asks how they work.

This adds GteMath, the SoftMath operations the GTE can do in hardware, and a
Utah teapot that uses it: 32 Newell bicubic patches tessellated at boot with
MVMVA, analytic normals from the patch tangents, nccs and ncct doing all the
shading, and a camera drifting along a Bezier path evaluated every frame.
Triangle, Circle and Cross switch a light each.

These are plain free functions because the measurements say they can be. On an
SCPH-5501 the GTE beats the software version at a batch of one for every
operation, 1.4x to 5.5x, even paying the matrix upload inside each call, so
there is no crossover to design around. Numbers on three consoles are in
src/mips/tests/gte-math-bench, coming separately. Each function documents what
it clobbers: the cross product eats the rotation matrix diagonal, which will
bite anyone who calls it inside a render loop.

The gouraud primitives gain getCommandWord and setColorAPacked so the GTE's
CODE field can be used for what it is for - fusing the GP0 command byte into
the colour it produces, which makes storing a lit vertex a word store.


rixnobis:
Edited one line of the body above. It promised `src/mips/tests/gte-math-bench, coming separately`, and that path is in neither repository now: the mips tree moved out in #2112, so the benchmark is in [pcsx-redux/nugget](https://github.com/pcsx-redux/nugget) as `tests/gte-math-bench` (0cb55e8a5), wired into the tests build and covered by that repo's Build workflow.

Its RESULTS.md names `fastNormalizeVec3`'s `x * 2` seed as the defect. nugget f1f109cf6 replaced that seed with a software leading-zero one, so those accuracy readings are fenced as historical. The cycle counts are untouched.

PR #2077, opened 2026-08-20
SPU rewrite

nicolasnoble:


rixnobis:
Pushed four commits to the head branch.

Three were sitting local since you answered their content: the key-on startup delay defaulting to 6 samples, a period the dummy audio driver can divide, and the accuracy suite wired into the build and CI.

The fourth came out of the pre-push scan, which refused the first attempt. `src/mips/tests/gpu-raster-phase1/raster-expected.h` cited a path on my own machine for facts a reader has no way to reach. `main` fixed that already; this branch had not caught up. It now points at the capture workflow directly below it. No functional change.

The branch is still conflicting against main.

rixnobis:
@Algebro7 this PR adds a `third_party/implot` submodule (https://github.com/epezent/implot.git), so `pcsx-redux-git` needs it in `source` and in the `git config submodule...url` rewrites. `aur-build` currently fails here on `Missing: third_party/implot/implot.h`. Same as #2128, it only matters once this merges.

PR #2102, opened 2026-09-16
Measure the LZCS to LZCR store delay on hardware

rixnobis:
New test under src/mips/tests/gte-latency-lzcs: write LZCS, pad with N nops, read LZCR, report the smallest N that returns the new count. Two cached, one uncached, identical on SCPH-1000, 1001, 5501 and 7001. Flat across 126 inputs covering every leading-zero and leading-one run at two popcounts, so the encoder is combinational. A too-early read gives a clean stale value, never garbage. PAL is not covered. Build with TYPE=ps-exe.

The torus example emits one nop there, so it seeds its inverse square root from the previous iteration's count. Not touched here.


src/mips/tests/gte-latency-lzcs/RESULTS.md
```diff
@@ -0,0 +1,148 @@
...
+Both attempts were leased, and both ended the same way. The submit
+client's stdout does not carry the terminal state; read it back from the
+lease:
+
+| Attempt | Ticket | Lease | Terminal state | Failure |
+|---|---|---|---|---|
+| 1 | `ticket_23c076b6-6e71-4779-a83a-094d367801b1` | `lease_8b314334` | FAILED 19:58:56Z | `lease_timeout: Lease hard deadline expired` |
+| 2 | `ticket_e6468924-40f4-4781-9e50-6d99359a21f5` | `lease_c13d532f` | FAILED 20:09:56Z | `lease_timeout: Lease hard deadline expired` |
+
+`scph9002-6` is the ONLY device in the pool advertising `flaky`,
```
nicolasnoble: Most of the narrated details here aren't really necessary. You're describing test fixture hiccups which aren't relevant to the actual results, aka the number of cycles required.

src/mips/tests/gte-latency-lzcs/RESULTS.md
```diff
@@ -0,0 +1,148 @@
...
+reports every single-nop read stale, on 126 inputs and four consoles. The
+two are consistent with a delay counted in clock cycles where surrounding
+real code sometimes fills the gap on its own: this probe is a straight
+line with interrupts masked and a warm icache, so nothing else can. The
+earlier harness was not re-run here, so that is a reading of the
+difference and not a measurement of it.
+
+## The 256 case, specifically
+
+`.siev` observed a case that appeared to work with one opcode and noted it
```
nicolasnoble: Same here. You're narrating a lot of who-said-what, but that's not really useful. You can be more concise here.

src/mips/tests/gte-latency-lzcs/RESULTS.md
rixnobis: Taken. The lease failures, ticket ids and device feature lists are gone, along
with the paragraph weighing up a third PAL lease. What survives from that block
is one sentence saying no PAL unit was measured, since that bounds the result.

src/mips/tests/gte-latency-lzcs/RESULTS.md
rixnobis: Also taken. The 256 section now states what the sweep shows and drops the
attributions. The file is down from 148 lines to 104.

nicolasnoble (dismissed):

src/mips/tests/gte-latency-lzcs/README.md
```diff
@@ -0,0 +1,91 @@
...
+  input-independent, so the smallest sufficient N is flat across inputs.
+  A shift-and-count iterates, so N would track the leading-bit count and
+  small inputs would settle sooner. A sweep over N alone returns one
+  number and cannot tell those apart.
+
+126 inputs cover **every** leading-zero run from 1 to 32 and **every**
+leading-one run from 1 to 32, each at two popcounts: a dense value
+(`0xffffffff >> k`) and a sparse one (a single bit, `1 << (31-k)`), so an
+implementation whose work tracks set bits is distinguishable from one
+whose work tracks the leading run.
```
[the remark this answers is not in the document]
rixnobis: Counted off `g_cases`: 126 rows, 124 distinct values. The collision is at run
31, where the dense and sparse forms are both `0x00000001` (and `0xfffffffe`
negated), plus the single row at run 32. Both documents now say rows where they
meant rows, and the README states the 124.

Nothing in either file claimed 128, so that part of the finding had no referent.

nicolasnoble (approved):

PR #2105, opened 2026-09-18
MDEC: a BS frame decoder for the console

rixnobis:
A BS frame decoder for the console: plain C in `src/mips/bsdec`, no supportpsx, no
allocation, host-buildable. Returns the run-level halfwords and the MDEC0 command word.

Every code in Sony's AC book is N zeros, a marker 1, then a fixed suffix keyed to N, so
the decode is one count-leading-zeros and one indexed load. The R3000A has no CLZ
opcode, hence GTE LZCS/LZCR. `gentable.py` generates the book from the encoder's table
and it fills all 269 slots, so there is no invalid-prefix arm.

Bit-identical to `DCT::fromContainer` on 7614 frames off five retail discs (168 v1, 7303
v2, 143 v3). `src/mips/tests/bsdec` runs it on hardware and self-tests the GTE count
first. `gentable.py` needs #2101 to regenerate; the generated header is committed, so
the library builds against main today.


src/mips/bsdec/bsdec.h
```diff
@@ -0,0 +1,139 @@
...
+ *    the read each need two dummy opcodes after them; cop2_put and cop2_get
+ *    carry those. One nop is NOT enough - the read comes back with the previous
+ *    write's answer about a third of the time, and a single isolated call passes
+ *    by luck, so only a loop self-test on real silicon exposes it.
+ */
+static inline uint32_t bsdecClz32(uint32_t v) {
+#ifdef __mips__
+    uint32_t r;
+    if ((int32_t)v < 0) return 0;
+    cop2_put(30, v);
```
nicolasnoble: The common/hardware/cop2.h file has defines for register names, which you should use instead of hard-coding their index here.

src/mips/tests/bsdec/bsdec.c
```diff
@@ -0,0 +1,227 @@
...
+ *    The host side diffs those against supportpsx' fromContainer, which is itself
+ *    graded against Sony's DecDCTvlc on the retail corpus. No MDEC and no DMA here
+ *    on purpose: this rig answers "does the bitstream decode correctly on real
+ *    silicon", and pulling the hardware in would fold two questions together.
+ *
+ * Runs identically under pcsx-redux and on a devkit. That is the point - the
+ * emulator need not model the missing interlock, so the emulator arm passing says
+ * nothing about hardware and the two arms have to be compared.
+ *
+ * ⛔ There is deliberately NO fallback frame. A rig that decodes something built
```
nicolasnoble: Please no special UTF-8/emojis... There's more down this file.

src/mips/bsdec/bsdec.h
```diff
@@ -0,0 +1,139 @@
...
+    cop2_get(31, r);
+    return r;
+#else
+    return v ? (uint32_t)__builtin_clz(v) : 32u;
```
nicolasnoble: **Potentially** we could ship the underlying libgcc-style function name in the same bulk of files we have in our crt0? This would (1) avoid the #ifdef __mips__ in this file, and (2) allow other usages of clz for other people.

rixnobis:
Addressed in `ed379a82b` and `2fe7b8665`.

The register names needed a header fix first: `cop2_put` stringifies its argument
unexpanded, so `cop2_put(COP2_LZCS, v)` assembled as `mtc2 $4,$COP2_LZCS`. All four
accessors now expand; numeric arguments unchanged, gte tests still build.

Took the `__clzsi2` suggestion. `__builtin_clz` lowers to `jal __clzsi2` on mips1 and
builds here are `-nostdlib`, so the builtin was an undefined reference tree-wide, which
takes a `volatile` input to see. `common/crt0/clz.c` answers it on LZCS/LZCR, weak,
beside `memory-c.c`. It needs CU2, which OpenBIOS sets and the retail BIOS left on via
its shell. The `#ifdef` is out of `bsdec.h` and the rig now self-tests the shared
intrinsic.

Glyphs gone; there were two.

CodeRabbit's boundary: writes are bounded on `count >= target` now. With slack in the
buffer the old capacity check could never fire and a block ran past the declared length.
Overlong streams get `BSDEC_OVERLONG`, and `bsdecUsable()` is the test, since the rig was
refusing every non-zero status. 10 frames reach that arm under a broken escape, all at
the declared length; 7614 retail frames still bit-identical.

spicyjpeg:
You may want to take a look at my own take on fast BS decoding, written in hand-optimized assembly and leveraging both LZCS/LZCR and the scratchpad ([implementation](https://github.com/spicyjpeg/573in1/blob/dev/src/common/mdec.s), [LUTs and initialization](https://github.com/spicyjpeg/573in1/blob/dev/src/common/mdec.cpp#L185-L319), [header](https://github.com/spicyjpeg/573in1/blob/dev/src/common/mdec.hpp#L79-L142)). I'd be interested in seeing how it compares to your implementation performance-wise on real hardware; mine seems to hover around 20-40% CPU usage during playback of a typical v2 or v3 video file encoded at 2x CD-ROM bitrate.

rixnobis:
Thanks, reading it now. The scratchpad half is the part I have nothing equivalent to: bsdec uses LZCS for the VLC length and then decodes straight into main RAM, so a block never stays resident.

State of my side, because a comparison is only worth having if both numbers are real. The decoder is measured under emulation only, 7614 retail frames bit-identical against `DecDCTvlc`. The console arm is written and unrun: `src/mips/tests/bsdec/`, with `-DBSDEC_TEST_BREAK_CLZ` as its negative control so a green cannot come from the harness never reaching the CLZ path. It is waiting on hardware time. I have no CPU figure to put beside your 20-40% and I am not going to estimate one.

Your number gives that run a target it did not have. Same series at 2x, or it is not a comparison.

rixnobis:
Read it. Your `mtc2 GTE_LZCS` at the top of the decode loop and the `mfc2 GTE_LZCR` at the head of the AC path sit two instructions apart, which is the minimum I measured on four consoles: 0/126 disagreements at two, and at one the read comes back with the previous write's answer about a third of the time. Your code is independent confirmation of a number I only had from my own harness.

It also turned up something in mine. My `__clzsi2` puts two dummy opcodes after the `mfc2` as well as after the `mtc2`, and only the write side was ever tested; yours leaves the ordinary single load-delay slot. I have no evidence the second one is needed, so that goes on the same harness next time I have a console out.

No CPU figure from me yet, because it would be the wrong number. `__builtin_clz` lowers to a call on mips1, so `bsdecFrame` spends a `jal`, three caller-saved spills, a callee that is four nops around one `mtc2`/`mfc2` pair, and three reloads on each AC coefficient: 17 instructions of plumbing against your two, and your LUTs sit in scratchpad while mine take main RAM latency per lookup. Against your 20-40% that mostly measures the O32 ABI. So the order here is silicon correctness first (that run has not happened either), then inlining the GTE, then cost. Inlining puts the nop count into the hot loop on the one axis Redux cannot test, and I would rather not have both unknowns in the same binary.

One scoping note so I do not mis-carry your figure: yours is resumable and mine decodes a frame per call, so 20-40% is occupancy during playback and anything I produce would be throughput.

rixnobis:
All four graded at source. Two fixed in `b93bb490`, two not.

**`decodeInto(i)` at line 482 is a hard compile error.** Two arms, same tree: as shipped, `make STREAM=... CAPTURE=3` exits 2 with `mdecplayer.cpp:482:28: error: no matching function for call to 'PlayScene::decodeInto(uint32_t&)'`; with the fix it exits 0 and links a 1482752-byte ps-exe. Nothing caught it because of the Makefile's own CI accommodation: `examples/` gets a bare `make all`, and an unset `STREAM` is a no-op there, so the capture path has never been compiled by anything.

**The MDEC re-seat was right too.** `m_needReset` is set inside `decodeInto` and consumed only in `frame()` after `waitUpload()`; the capture block `pcsx_exit`s before reaching that, so a capture of frame n > 0 fed the next command into a half-consumed MDEC. The loop now re-seats between decodes and exits 24 if `mdecReset()` reports a bad DMA.

**On the frame interval, `m_shownAt` is declared, zeroed in `start()`, assigned after every successful decode, and read nowhere.** Somebody meant to write that comparison and never did. I have not fixed it: capping playback changes what the profiling build measures, that harness is going to @nicolasnoble with the decoder, and whether this example is my lane depends on a scope question I have open with him. It stays on my list.

**`refillBytes` is wrong in both ways you name.** It increments once per pull, every pull is a halfword pair since the bit-bucket change, and it counts the synthetic zeros fed past the payload, so it reads about double its documented meaning with a contaminated tail. Leaving it alone: `bsdec.c` is @nicolasnoble's as of tonight, and a counter rename is a bad thing to collide with. The published hblank and cycle figures come from the counter-2 arms and do not rest on it.

rixnobis:
Moved to pcsx-redux/nugget#9. #2112 took `src/mips` out of this repo and mounts it as a submodule, so this cannot land here any more.

The 21 commits are replayed intact with `format-patch --relative`. One commit is new on top: `tests/bsdec` reached `third_party` four levels up from this repo's root, and in nugget it is two.

Built in `ghcr.io/grumpycoders/pcsx-redux-build:latest` before opening it - `tests/bsdec`, `psyqo/examples/bsdecbench` and `psyqo/examples/mdecplayer` each exit 0.

src/mips/bsdec/bsdec.h
rixnobis: Done in the move to nugget: bsdec.h no longer touches COP2 registers. The one place left that named them by number, the broken-clz control in tests/bsdec/bsdec.c, is pcsx-redux/nugget#18.

src/mips/tests/bsdec/bsdec.c
rixnobis: Done, tests/bsdec/bsdec.c has no non-ASCII left on nugget main. Sorry for going silent on these.

src/mips/bsdec/bsdec.h
rixnobis: Done: `__clzsi2` is in common/crt0/clz.c on nugget, GTE-backed, and bsdec uses it without the `#ifdef __mips__`.

PR #2110, opened 2026-09-19
Quote the credential URLs in the export push steps

rixnobis:
Shellcheck SC2086 from the CodeRabbit pass on #2109: the five `https://rixnobis:$PAT@...` URLs in the export steps were unquoted. Quoted them.

This is hygiene, not a bug fix. A GitHub token carries no whitespace or glob characters, so word splitting could not fire on a well-formed secret; quoting just keeps the next lint pass quiet.

Noticed while reading and left alone: the doxygen step sets `user.email "pixel@nobis-crew.org."` with a trailing period. Your call whether that matters on the gh_pages commits.


nicolasnoble:
Hah, we should also fix this typo.

rixnobis:
The two reds here are CodeSee refusing a fork checkout and an aur-build run from before the AUR package caught up with the submodule move; neither is required.

PR #2112, opened 2026-09-20
Stop the mips export force-pushing nugget's main

rixnobis:
Taking ownership of nugget. It is mine now and I am not handing it back.

`src/mips` is going to live there as upstream with redux consuming it, so this repo does not get to force-push over nugget any more. That push is the thing that would flatten nugget's own commits on every unrelated push to main here, including ones that never touch `src/mips`.

Mechanically it is one step. The job still filters, runs doxygen and force-pushes `gh_pages`, so the documentation site is unaffected. The `extraheader` unset moved into the documentation push because that step needed it and did not have it: `actions/checkout` sets `http.https://github.com/.extraheader`, which applies to any github.com URL and overrides the inline credentials in the explicit push URL.


rixnobis:
This branch breaks the support export on merge. No check here can catch it.

`filter-support` keeps `src/mips/common` with `find src/mips -depth -type f -not -path src/mips/common/*`. Once `src/mips` is a gitlink that walk finds nothing, so `pcsx-redux/support` loses `mips/common` on the first push to main after this lands. `export-support` only runs on push to main, never on pull requests, so nothing here will ever go red over it.

The fix agreed with the maintainer is to give `mips/common` to support, since it is already the shared layer: `src/core/r3000a.h:35` includes `mips/common/util/mips.hh` and eight host files under `src/` and `tools/` include something under `mips/`. `tools/exe2iso` and `tools/ps1-packer` move into the support export at the same time, which is what lets nugget build an ISO without reaching back into this repo. That work is not in this branch.

Also open: #2113 for the nix build, which is wider than filed since `withOpenbios = false` breaks too, and the AUR package, whose PKGBUILD lives at `aur.archlinux.org/pcsx-redux-git.git` and hardcodes the old submodule paths.

nicolasnoble (approved):

